PR #16, opened 2019-08-12
Fix account creation from raw public/private keys

Vervious:
The previous constructor was broken and would throw an exception.


Vervious:
There is a strong argument for removing this constructor entirely, as we can also build the account from just the seed.

PR #18, opened 2019-08-12
Add signing and verifying for arbitrary bytes

mjiang102628:


src/main/java/com/algorand/algosdk/account/Account.java
```diff
@@ -273,6 +274,29 @@ public Signature signBytes(byte[] bytes) throws NoSuchAlgorithmException {
         }
     }
 
+    /**
+     * Sign the given bytes, and wrap in Signature.
```
Vervious: Worth documenting that we prepend the message with a prefix for domain separation / security

src/main/java/com/algorand/algosdk/account/Account.java
```diff
@@ -273,6 +274,29 @@ public Signature signBytes(byte[] bytes) throws NoSuchAlgorithmException {
...
+     * @param bytes the data to sign
+     * @return a signature
```
Vervious: for consistency (in the java interface), I'd prefer we return a Signature object (and the client can call sig.getBytes). Is there a strong argument for returning raw bytes?

src/main/java/com/algorand/algosdk/account/Account.java
```diff
@@ -273,6 +274,29 @@ public Signature signBytes(byte[] bytes) throws NoSuchAlgorithmException {
...
+        try {
+            CryptoProvider.setupIfNeeded();
+            java.security.Signature signer = java.security.Signature.getInstance(SIGN_ALGO);
+            signer.initSign(this.privateKeyPair.getPrivate());
+            // prepend hashable prefix
+            byte[] prefixBytes = new byte[bytes.length + BYTES_SIGN_PREFIX.length];
+            System.arraycopy(BYTES_SIGN_PREFIX, 0, prefixBytes, 0, BYTES_SIGN_PREFIX.length);
+            System.arraycopy(bytes, 0, prefixBytes, BYTES_SIGN_PREFIX.length, bytes.length);
+            // sign
+            signer.update(prefixBytes);
```
Vervious: call rawSignBytes in here to prevent code duplication

src/main/java/com/algorand/algosdk/crypto/Address.java
```diff
@@ -110,6 +128,38 @@ public String encodeAsString() throws NoSuchAlgorithmException {
...
+     * verifyBytes verifies that the signature for the message is valid for the public key
+     * @param message the message that was signed
+     * @param signature
+     * @return boolean; true if the signature is valid
+     */
+    public boolean verifyBytes(byte[] message, byte[] signature) throws NoSuchAlgorithmException, InvalidKeySpecException, InvalidKeyException, SignatureException {
+        CryptoProvider.setupIfNeeded();
+        X509EncodedKeySpec pkS;
+        try {
+            // Wrap both keys in ASN.1 format.
```
Vervious: nit: "both keys" --> "public key"

src/main/java/com/algorand/algosdk/crypto/Address.java
```diff
@@ -110,6 +128,38 @@ public String encodeAsString() throws NoSuchAlgorithmException {
         return res;
     }
 
+    /**
+     * verifyBytes verifies that the signature for the message is valid for the public key
```
Vervious: Also worth documenting the MX prefix here, lest users get confused and try to verify generic signatures

src/main/java/com/algorand/algosdk/crypto/Address.java
```diff
@@ -110,6 +128,38 @@ public String encodeAsString() throws NoSuchAlgorithmException {
...
+        // create public key
+        KeyFactory kf = KeyFactory.getInstance(KEY_ALGO);
+        PublicKey pk = kf.generatePublic(pkS);
+        
+        // verify signature
+        java.security.Signature sig = java.security.Signature.getInstance(SIGN_ALGO);
+        sig.initVerify(pk);
+        sig.update(prefixBytes);
+        return sig.verify(signature);
+    }
```
Vervious: Can you add a unit test? (Unless there's a cucumber test already that tests at this level of granularity?)

src/main/java/com/algorand/algosdk/crypto/Address.java
```diff
@@ -110,6 +128,38 @@ public String encodeAsString() throws NoSuchAlgorithmException {
         return res;
     }
 
+    /**
+     * verifyBytes verifies that the signature for the message is valid for the public key
+     * @param message the message that was signed
+     * @param signature
+     * @return boolean; true if the signature is valid
+     */
+    public boolean verifyBytes(byte[] message, byte[] signature) throws NoSuchAlgorithmException, InvalidKeySpecException, InvalidKeyException, SignatureException {
```
Vervious: A second benefit of wrapping `byte[] signature` in a `Signature` is you get length checking for free

src/main/java/com/algorand/algosdk/crypto/Address.java
mjiang102628: There's a positive and negative test in testSignBytes for verifying.

src/test/java/com/algorand/algosdk/account/TestAccount.java
```diff
@@ -204,4 +206,17 @@ public void testMergeMultisigTransaction() throws Exception {
         Assert.assertArrayEquals(a, b);
         Assert.assertArrayEquals(expected, a);
     }
+
+    @Test
+    public void testSignBytes() throws Exception {
```
Vervious: can you also add a version of this test that feeds in a golden value?

Vervious:
If this change is versioned (across the repositories), or if we need to push a maven update, we'll need to update the version numbers in `README`, `pom.xml`, `pom.xml.template`, and the changelog.

PR #19, opened 2019-08-13
Add functions to convert algos to and from microalgos

mjiang102628:


src/main/java/com/algorand/algosdk/util/AlgoConverter.java
```diff
@@ -0,0 +1,31 @@
+package com.algorand.algosdk.util;
+
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+
+public class AlgoConverter {
+
+    private static BigDecimal ALGOS_TO_MICROALGOS_RATIO = BigDecimal.valueOf(1000000L);
```
rotemh: if there is no centralized configs params class in the package, consider making it `final` as well. If there is one, move it there.

PR #20, opened 2019-08-29
replace PublicKey with Ed25519PublicKey

mjiang102628:
- remove duplicate classes; replace with equivalent existing ones


mjiang102628:
Using Ed25519PublicKey instead of adding bytes field to PublicKey class, as doing so would basically make these two classes duplicates, and still would not avoid having to do extra work for serialization.

src/main/java/com/algorand/algosdk/kmd/client/api/DefaultApi.java
```diff
@@ -1669,14 +1669,14 @@ public void onRequestProgress(long bytesWritten, long contentLength, boolean don
         return call;
     }
     /**
-     * Build call for listMultisg
+     * Build call for listMultisig
```
EvanJRichard: All these typo fixes are much appreciated/good catch

PR #22, opened 2019-09-04
No need to cast Authentication instance from ApiClient::getAuthentication

gubatron:
Both implementations of getAuthentication(...) on the ApiClient classes (algorand, kmdApi) are now generic.

The methods are able to deduce the return type based on the receiving variable, no need to be doing the cast every time we invoke `client.getAuthentication("...")`

Cheers!


gubatron:
CLA agreeement signed

PR #24, opened 2019-09-13
Draft: Update Transaction class to support asset creation.

EvanJRichard:


EvanJRichard:
Closing this - looks like we get it for free through asset config?

PR #30, opened 2019-10-01
Shant/java example readme changes

algonautshant:
1. The algod network address and token are hard-coded in the Java demo
program. Changing the program to expect these values as parameters.

2. Adding standard output messages to expect an overspending exception
before the error message is printed out.

3. algod.net address does not specify the http protocol. Adding logic
to the example program to amend the protocol if not specified in the
address passed as an argument.

4. Updating the README.md file to describe how to add the algod
network address and token parameters.

5. Adding cleanupDaemonThreads=false flag to prevent maven from
throwing java.lang.IllegalThreadStateException

6. Removing unused import from the java file.

### DIFF
--- a/examples/src/main/java/com/algorand/algosdk/example/Main.java
+++ b/examples/src/main/java/com/algorand/algosdk/example/Main.java
@@ -19,15 +19,29 @@ import com.algorand.algosdk.transaction.SignedTransaction;
 import com.algorand.algosdk.transaction.Transaction;
 import com.algorand.algosdk.util.Encoder;
 
-import java.security.Security;
+import com.squareup.okhttp.HttpUrl;
+
 import java.math.BigInteger;
+
 
 
 public class Main {
 
     public static void main(String args[]) throws Exception {
-        final String ALGOD_API_ADDR = "http://localhost:8080";
-        final String ALGOD_API_TOKEN = "b3ff7b3765bf33c7557a2cd334238eca07578e3ce266cb03fad93d12aab0622e";
+        if (args.length != 2) {
+            System.err.println("Required parameters: ALGOD_API_ADDR ALGOD_API_TOKEN");
+            System.err.println("The parameter values can be found in algod.net and algod.token files within the data directory of your algod install.");
+            System.exit(1);
+        }
+
+        // If the protocol is not specified in the address, http is added.
+        String algodApiAddrTmp = args[0];
+        if (HttpUrl.parse(algodApiAddrTmp) == null) {
+            algodApiAddrTmp = "http://" + algodApiAddrTmp;
+        }
+
+        final String ALGOD_API_ADDR = algodApiAddrTmp;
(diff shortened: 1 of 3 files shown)


examples/src/main/java/com/algorand/algosdk/example/Main.java
```diff
@@ -19,15 +19,29 @@
...
 public class Main {
 
     public static void main(String args[]) throws Exception {
-        final String ALGOD_API_ADDR = "http://localhost:8080";
-        final String ALGOD_API_TOKEN = "b3ff7b3765bf33c7557a2cd334238eca07578e3ce266cb03fad93d12aab0622e";
+        if (args.length != 2) {
+            System.err.println("Required parameters: ALGOD_API_ADDR ALGOD_API_TOKEN");
+            System.err.println("The parameter values can be found in algod.net and algod.token files within the data directory of your algod install.");
+            System.exit(1);
+        }
```
pzbitskiy: put examples for parameters to give people an idea what to put to the command line

examples/src/main/java/com/algorand/algosdk/example/Main.java
```diff
@@ -19,15 +19,29 @@
...
+
+        // If the protocol is not specified in the address, http is added.
+        String algodApiAddrTmp = args[0];
+        if (HttpUrl.parse(algodApiAddrTmp) == null) {
```
pzbitskiy: I do not think we want a new 3rd party dependency only for protocol checking. Let's keep it lean and use startsWith()

algonautshant: This is not necessarily a new dependency. 
It is already used in src/main/java/com/algorand/algosdk/algod/client/ApiClient.java and the example eventually depends on it. 
What do you think?

examples/src/main/java/com/algorand/algosdk/example/Main.java
algonautshant: I was hoping that the README file was fulfilling this request. 
The maven command in the README file gives an argument template: 
127.0.0.1:8080 X-Algo-API-Token
Putting random characters as an example for the token may not be very helpful.

examples/src/main/java/com/algorand/algosdk/example/Main.java
pzbitskiy: Did not know that, then it's ok

examples/src/main/java/com/algorand/algosdk/example/Main.java
pzbitskiy: maybe change README to ${X_Algo_API_Token} then?

examples/src/main/java/com/algorand/algosdk/example/Main.java
algonautshant: I think your point is still valid. bazel complained! I'm taking your suggestion.

PR #32, opened 2019-10-05
[Important] - Change the word setupIfNeeded to setup

rotemh:
-- This is an important fix. 

Apparently, there was a typo in one of the Mnemonic words. Luckily, the fix is simple and there are no severe consequences. When one upgrades to this patch, there is a need to let the users know that if they have the word "setupIfNeeded" in their passphrase, they should change it to "setup". Everything else remains the same and it doesn't affect their private keys.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+
+# 1.1.6
+## Changed
+- IMPORTANT - This version modifies one of the mnemonic words. Please let your users to know that if they have the word "setupIfNeeded" in their mmnemonic, they should change it to "setup". Everything else remains the same. 
 # 1.1.5
 ## Added
 - signing and verifying signatures for arbitrary bytes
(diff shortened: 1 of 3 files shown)


rotemh:
I'm thinking of adding another notice in a more prominent location like the README file. Thoughts are welcome

shaih:
I agree about the prominent notice. In addition to github, this change should be communicated in every channel possible to the users, a news item, a blog post, direct email where possible, etc.

rotemh:
> The only suggestion to throw on "setupIfNeeded" in mnemonic parsing and request to change. Then to remove this after two releases.

Do you suggest to still accept the word for the next 2 versions?

pzbitskiy:
To accept the mnemonic by Account constructor but raise an exception saying
something like "replace 'setupIfNeeded' by 'setup' word". Simply because
most people do not read the documentation and changelogs. But I just
realized that the problem is even worse for a case of the mnemonic creation
by Java SDK and then importing with `goal` command.

PR #48, opened 2019-10-30
Java SDK changes to run full assets cucumber tests

algonautshant:
Enhanced ApiException in kmd to report response body in exception.
Added a utility method to Transaction to extract com.algorand.algosdk.algod.client.model.AssetParams from it.


algonautshant:
Updates to java sdk to reflect assetid changes 
Removing creator field.
Adding classes AssetHolding and Participation

algonautshant:
Updated golden in the tests.

src/main/java/com/algorand/algosdk/account/Account.java
```diff
@@ -207,12 +207,6 @@ static public Transaction transactionWithSuggestedFeePerByte(Transaction copyTx,
             case KeyRegistration:
                 return new Transaction(copyTx.sender, newFee, copyTx.firstValid, copyTx.lastValid, copyTx.note, copyTx.genesisID, copyTx.genesisHash,
                         copyTx.votePK, copyTx.selectionPK, copyTx.voteFirst, copyTx.voteLast, copyTx.voteKeyDilution);
-            case AssetConfig:
-                return new Transaction(copyTx.sender, newFee, copyTx.firstValid, copyTx.lastValid, copyTx.note, copyTx.genesisID, copyTx.genesisHash,
-                        copyTx.assetID, copyTx.assetParams);
-            case AssetFreeze:
-                return new Transaction(copyTx.sender, newFee, copyTx.firstValid, copyTx.lastValid, copyTx.note, copyTx.genesisID, copyTx.genesisHash,
-                        copyTx.assetFreezeID, copyTx.freezeTarget, copyTx.freezeState);
```
EvanJRichard: I think these need to stay, don't they? Or is that not true, since the function is deprecated?

src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -574,8 +625,7 @@ public static Transaction createTransferAssetTransaction(// AssetTransaction
                 note, //note
                 genesisID, // gen
                 genesisHash, // gh
-                assetCreator, // c
-                assetIndex); // i
+                assetIndex); //
```
EvanJRichard: do we want the empty comment here?

src/main/java/com/algorand/algosdk/account/Account.java
algonautshant: function is deprecated. It should not support new features.

src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -498,21 +524,19 @@ private Transaction(// Asset xfer transaction
      * @param genesisID corresponds to the id of the network
      * @param genesisHash corresponds to the base64-encoded hash of the genesis
      * of the network
-     * @param assetCreator is the address of the asset creator
      * @param assetIndex is the asset index
      **/
-    public static Transaction createAcceptAssetTransaction( //AssetTransaction
+    public static Transaction assetAcceptanceTransaction( //AssetTransaction
```
rotemh: Since it's static, what do you think of naming it something like `CreateAssetAcceptanceTransaction`? same for the other onces

algonautshant: How about this proposal: 
AssetCreateTransaction -> createAssetCreationTransaction or createAssetCreateTransaction 
assetConfigurationTransaction -> createAssetConfigurationTransaction or createAssetConfigureTransaction
assetAcceptanceTransaction -> createAssetAcceptanceTransaction or createAssetAcceptTransaction
assetDestroyTransaction -> createAssetDestroyTransaction
assetFreezeTransaction -> createAssetFreezeTransaction
assetRevokeTransaction -> createAssetRevokeTransaction
assetTransferTransaction -> createAssetTransferTransaction

rotemh: Well, the first ones are Objects - they should not get a verb. 
The last ones are static functions that return the object. I think that they should get the verb.

PR #51, opened 2019-11-12
Java sdk lease

algonautshant:
Adding lease to java SDK transaction with test.
Also added a missing pay transaction.

### DIFF
--- a/src/main/java/com/algorand/algosdk/transaction/Transaction.java
+++ b/src/main/java/com/algorand/algosdk/transaction/Transaction.java
@@ -26,7 +26,7 @@ import java.util.Map;
 @JsonInclude(JsonInclude.Include.NON_DEFAULT)
 public class Transaction implements Serializable {
     private static final byte[] TX_SIGN_PREFIX = ("TX").getBytes(StandardCharsets.UTF_8);
-
+    private static final int LEASE_LENGTH = 32;
     @JsonProperty("type")
     public Type type = Type.Default;
 
@@ -739,7 +739,7 @@ public class Transaction implements Serializable {
      * @param lease 32 byte lease
      **/
     public void setLease(byte[] lease) {
-        if (lease.length != 32 && lease.length != 0) {
+        if (lease.length != LEASE_LENGTH && lease.length != 0) {
             throw new RuntimeException("The lease should be an empty array or a 32 byte array.");
         }
         this.lease = lease;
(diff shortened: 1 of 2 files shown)


src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -461,6 +466,7 @@ private Transaction(
         if (note != null) this.note = note;
         if (genesisID != null) this.genesisID = genesisID;
         if (genesisHash != null) this.genesisHash = genesisHash;
+        if (lease != null) this.lease = lease;
```
EvanJRichard: Do we want a length check here, as well, or is that redundant?

rotemh:
Is the only way to add a lease value is after creating the transaction?

src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -723,6 +729,22 @@ public static Transaction createAssetTransferTransaction(// AssetTransaction
...
+     * is nonzero, then once the transaction is confirmed, it acquires
+     * the lease identified by the (Sender, Lease) pair of the
+     * transaction until the LastValid round passes.  While this
+     * transaction possesses the lease, no other transaction
+     * specifying this lease can be confirmed. 
+     * The Size is fixed at 32 bytes. 
+     * @param lease 32 byte lease
+     **/
+    public void setLease(byte[] lease) {
+        if (lease.length != 32 && lease.length != 0) {
```
rotemh: Consider making it a constant to avoid magic numbers

src/main/java/com/algorand/algosdk/transaction/Transaction.java
algonautshant: Unfortunately, all the members are public, including lease. Which means, any code can set the lease value to anything and cannot be checked. 

Setting it properly though setLease, will check the value. 

Checking the assignment here is not necessary, because it is a private constructor, called from a private method, used for parsing external transactions, which is assumed to be well-formed and not to violate the lease size.

src/main/java/com/algorand/algosdk/transaction/Transaction.java
algonautshant: Fixed!

algonautshant:
> Is the only way to add a lease value is after creating the transaction?

Yes.

PR #54, opened 2019-11-15
java SDK model update: transaction results

algonautshant:
The updates include: transaction result class in algod.client.model
Added new file:
 src/main/java/com/algorand/algosdk/algod/client/model/TransactionResults.java


rotemh:
Seems like there are many unrelated updates here (All the asset types). Is that another PR that got mixed in?

PR #70, opened 2019-12-05
add decimal field to AssetParams

winder:


src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -901,6 +903,14 @@ public boolean equals(Object o) {
         @JsonProperty("t")
         public BigInteger assetTotal = BigInteger.valueOf(0);
 
+        // Decimals specifies the number of digits to display after the decimal
+        // place when displaying this asset. A value of 0 represents an asset
+        // that is not divisible, a value of 1 represents an asset divisible
+        // into tenths, and so on. This value must be between 0 and 19
+        // (inclusive).
+        @JsonProperty("dc")
+        public BigInteger assetDecimals = BigInteger.valueOf(0);
```
algonautshant: assetDecimals type better be byte, short or int, and not BigInteger. It holds only 19 values, and it is implemented as 32 bit int on the server side. 
Primitive types are easier to handle in the code than BigInteger. Let us not add more of it. At some point, it should all go away.

src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -950,6 +961,7 @@ public AssetParams(
 			   Address freeze,
 			   Address clawback) {
             if(assetTotal != null) this.assetTotal = assetTotal;
+            if(assetDecimals != null) this.assetDecimals = assetDecimals;
```
algonautshant: null check won't be needed here once assetDecimals is converted to a primitive data type.

rotemh: Please check that the number is indeed between 0 and 19

rotemh:
> 2. How can half an asset be transferred with the java api? Asset transfer transaction still takes integer value with no decimal part. Should there be changes to asset transfer api as well?
Transfers are always specified but the base unit. So say ALGO was an asset. you should simply transfer 500000 base units. I don't think we should go into decimal conversions.

winder:
@algonautshant thanks! I didn't realize there was no need for BigInteger!

winder:
@rotemh @algonautshant I believe this is ready to go. I'll make a followup PR to finish removing Bazel.

src/main/java/com/algorand/algosdk/transaction/Transaction.java
rotemh: what about this check?

winder:
@rotemh added missing field validation.

src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -951,14 +962,35 @@ public AssetParams(
...
-            if(assetName != null) this.assetName = assetName;
-            if(url != null) this.url = url;
-            if(metadataHash != null) this.metadataHash = metadataHash;
             if(manager != null) this.assetManager = manager;
             if(reserve != null) this.assetReserve = reserve;
             if(freeze != null) this.assetFreeze = freeze;
             if(clawback != null) this.assetClawback = clawback;
+
+            if(assetDecimals != null) {
+                if (assetDecimals < 0 || assetDecimals > 19) throw new RuntimeException("assetDecimals cannot be greater than 19");
```
rotemh: nit - add that they number should be also positive to the comment. Otherwise, if the user enters a negative number for some reason, s/he will get the wrong error.

PR #74, opened 2019-12-13
Added check for valid Base64 encoding of asset metadataHash.

egieseke:
Added check for base64 encoding for asset metadataHash.

### DIFF
--- a/src/main/java/com/algorand/algosdk/transaction/Transaction.java
+++ b/src/main/java/com/algorand/algosdk/transaction/Transaction.java
@@ -995,7 +995,7 @@ public class Transaction implements Serializable {
             }
 
             if(metadataHash != null) {
-                if (!Base64.isBase64(metadataHash)) throw new RuntimeException("asset metadataHash is not base64 encoded");
+                if (!Base64.isBase64(metadataHash)) throw new RuntimeException("asset metadataHash '" + new String(metadataHash) + "' is not base64 encoded");
                 this.metadataHash = metadataHash;
             }
         }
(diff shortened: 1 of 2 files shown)


src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -992,6 +993,11 @@ public AssetParams(
                 if (metadataHash.length > 32) throw new RuntimeException("asset metadataHash cannot be greater than 32 bytes");
                 this.metadataHash = metadataHash;
             }
+
+            if(metadataHash != null) {
+                if (!Base64.isBase64(metadataHash)) throw new RuntimeException("asset metadataHash is not base64 encoded");
```
algonautshant: Please also add the metadataHash to the exception message. It will be easier to identify with metadataHash is not well-formed.

winder: please move this into the previous `metadataHash != null` check

winder:
Does this really need to be bsae64 encoded? The other SDKs just use a string?

winder:
> It is not a good idea to throw RuntimeException because java does not require methods to catch runtime exceptions. It is more for unexpected things. In this case, it is better to make an algorand java sdk specific exception and throw it. This way, a method calling AssetParams, can be aware of the exception, and decide to handle it or throw it.
> I should have mentioned this in #70 when the other runtime exceptions were added.

At the moment `RuntimeException` is widely used in this SDK, and I don't think we should break backwards compatibility for this change by switching to a checked exception.

src/main/java/com/algorand/algosdk/transaction/Transaction.java
egieseke: Added metadataHash to the error message and combined blocks together.

PR #79, opened 2020-01-06
Add multiple parameter value classes

winder:
Created a number of ParameterValue types to make the contract implementation slightly less verbose.

### DIFF
--- a/src/main/java/com/algorand/algosdk/templates/ContractTemplate.java
+++ b/src/main/java/com/algorand/algosdk/templates/ContractTemplate.java
@@ -17,63 +17,66 @@ public class ContractTemplate {
 	/**
 	 * Values are appended to a program at specific offsets. Depending on the type the source template has placeholders
 	 * of differing sizes, this offset is returned by the "updateProgram" method and is used to traverse the template.
+	 *
+	 * Note: Placeholder sizes may differ between SDK implementations.
      *
-	 * Offsets are the following:
-	 * 		BASE64 : placeholder  2 bytes
-	 * 		ADDRESS: placeholder 32 bytes
-	 * 		INT    : placeholder  1 byte
+	 * Placeholder sizes are the following:
+	 * 		BASE64 : 2 bytes
+	 * 		ADDRESS: 32 bytes
+	 * 		INT    : 1 byte
 	 */
 	abstract static class ParameterValue {
-		abstract public int updateProgram(ArrayList<Byte> program);
+        private final byte[] value;
+
+        protected ParameterValue(byte[] value) {
+        	this.value = value;
+		}
+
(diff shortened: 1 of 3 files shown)


src/main/java/com/algorand/algosdk/templates/ContractTemplate.java
```diff
@@ -14,44 +14,66 @@
 
 public class ContractTemplate {
 
-	public enum DataType {
-		BASE64, ADDRESS, INT;
+	/**
+	 * Values are appended to a program at specific offsets. Depending on the type the source template has placeholders
```
algonautshant: Do you want to mention here something that these are not cross-SDK compatible? That each binary works only within the SDK it was designed for, and that the placeholder sizes are also SDK specific?

src/main/java/com/algorand/algosdk/templates/ContractTemplate.java
```diff
@@ -14,44 +14,66 @@
...
+	 * of differing sizes, this offset is returned by the "updateProgram" method and is used to traverse the template.
+     *
+	 * Offsets are the following:
```
algonautshant: I do not think the use of the term "Offsets" here is correct. The following of the placeholder sizes, and not offsets.

src/main/java/com/algorand/algosdk/templates/ContractTemplate.java
winder: Good point, clarification added.

src/main/java/com/algorand/algosdk/templates/ContractTemplate.java
winder: Yes, you're right. I've tried to clarify this.

winder:
@algonautshant thanks for the feedback, I've made changes to address the problems.

src/main/java/com/algorand/algosdk/templates/ContractTemplate.java
```diff
@@ -155,7 +158,11 @@ protected static ContractTemplate inject(byte [] program, int [] offsets, Parame
 			if (paramIdx < offsets.length && offsets[paramIdx] == progIdx) {
 				ParameterValue value = values[paramIdx];
 				++paramIdx;
-				progIdx += value.updateProgram(updatedProgram);
+				for (byte b : value.toBytes()) {
+					updatedProgram.add(b);
+				}
+				progIdx += value.placeholderSize();
+				//progIdx += value.appendToProgram(updatedProgram);
```
algonautshant: Is this commented line needed?

winder: No, removed.

PR #80, opened 2020-01-07
PeriodicPayment template

winder:


src/test/java/com/algorand/algosdk/templates/TestTemplates.java
```diff
@@ -48,14 +43,31 @@ public void testSplit() throws Exception {
...
+
+		assertThat(Encoder.encodeToBase64(result.program)).isEqualTo(goldenProgram);
+		assertThat(result.address.toString()).isEqualTo(goldenAddress);
+	}
+
+	@Test
+	public void testPeriodicPayment() throws Exception {
+		String goldenAddress = "IJPPJDULMZNQXBIGCLCWVP5P4VD6U564BHC7C5CMZB2FVXKGZSIVPJFPT4";
+		String goldenProgram = "ASAHAQoLAOcHkE7AxAcmAiB/g7Flf/H8U7ktwYFIodZd/C1LH6PWdyhK3dIAEm2QaSD+vKC7FEpaTqe0OKRoGsgObKEFvLYH/FZTJclWlfaiEzEQIhIxASMOEDECJBglEhAxBCEEMQIIEhAxBigSEDEJMgMSMQcpEhAxCCEFEhAxCSkSMQcyAxIQMQIhBg0QMQglEhAREA==";
+		String goldenWithdrawlTransaction = "gqRsc2lngaFsxJcBIAcBCgsA5weQTsDEByYCIH+DsWV/8fxTuS3BgUih1l38LUsfo9Z3KErd0gASbZBpIP68oLsUSlpOp7Q4pGgayA5soQW8tgf8VlMlyVaV9qITMRAiEjEBIw4QMQIkGCUSEDEEIQQxAggSEDEGKBIQMQkyAxIxBykSEDEIIQUSEDEJKRIxBzIDEhAxAiEGDRAxCCUSEBEQo3R4bomjYW10zScQo2ZlZQqiZnbNBLqiZ2jEIH+DsWV/8fxTuS3BgUih1l38LUsfo9Z3KErd0gASbZBpomx2zQihomx4xCB/g7Flf/H8U7ktwYFIodZd/C1LH6PWdyhK3dIAEm2QaaNyY3bEIP68oLsUSlpOp7Q4pGgayA5soQW8tgf8VlMlyVaV9qITo3NuZMQgQl70jotmWwuFBhLFar+v5Ufqd9wJxfF0TMh0Wt1GzJGkdHlwZaNwYXk=";
```
winder: The withdrawal stx golden needs to be verified.

src/main/java/com/algorand/algosdk/logic/Logic.java
```diff
@@ -41,17 +41,90 @@
...
+                s += 7;
+            }
+            return new VarintResult();
+        }
+    }
+
     private static LangSpec langSpec;
     private static Operation[] opcodes;
 
+    public static boolean checkProgram(byte[] program, ArrayList<byte[]> args) throws IOException {
```
algonautshant: Since checkProgram is only used in the test, it is better if it is a test helper method. 
It is not needed here.

src/main/java/com/algorand/algosdk/templates/PeriodicPayment.java
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * Return the withdrawal transaction to be sent to the network.
+     *
+     * @param contract contract containing information, this should be provided by the payer.
+     * @param firstValid first round the transaction should be valid.
+     * @param genesisHash genesis hash in base64.
+     * @return Signed withdrawal transaction.
+     */
+    public static SignedTransaction MakeWithdrawalTransaction(final byte[] contract, final int firstValid, final Digest genesisHash) throws IOException, NoSuchAlgorithmException {
```
algonautshant: Why take the contract as byte[]? 
It will be nicer to have it in more specific data type, which is ContractTemplate.

src/main/java/com/algorand/algosdk/logic/Logic.java
winder: This method isn't actually new, I must have moved it from elsewhere in the code. It's used by the LogicSignature class.

src/main/java/com/algorand/algosdk/templates/PeriodicPayment.java
winder: Good idea, I'll also need to enhance `ContractTemplate` to make it easier to initialize but that will be another good change.

src/main/java/com/algorand/algosdk/logic/Logic.java
pzbitskiy: checkProgram is used on LogicSig creation when a program supplied by a user, so it is not test-only helper

src/main/java/com/algorand/algosdk/logic/Logic.java
```diff
@@ -41,17 +41,90 @@
...
+        public final int size;
+        public final ArrayList<byte[]> results;
+
+        ByteConstBlock(int size, ArrayList<byte[]> results) {
+            this.size = size;
+            this.results = results;
+        }
+    }
+
+    protected static class Uvarint {
```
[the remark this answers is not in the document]
winder: Thanks, I didn't notice that

winder:
@algonautshant thanks for the suggestions, they should all be addressed now.

PR #83, opened 2020-01-14
Refactoring to better implement DynamicFee Template

winder:
This is refactoring / cleanup work done while implementing the DynamicFee template. There are no functional changes here.


src/main/java/com/algorand/algosdk/crypto/LogicsigSignature.java
```diff
@@ -61,6 +61,10 @@ public LogicsigSignature(
         this.msig = msig;
     }
 
+    public LogicsigSignature(byte[] logic) {
```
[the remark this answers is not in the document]
algonautshant: Also, the definition of the class LogicsigSignature does not have a description about what this class is. 
It will be very helpful to follow with the logic in these classes if the LogicSignature class is described.

src/main/java/com/algorand/algosdk/templates/PeriodicPayment.java
```diff
@@ -44,11 +44,7 @@ public static ContractTemplate MakePeriodicPayment(final String receiver, final
     /**
      * Need a way to specify the lease for testing.
      */
-    protected static ContractTemplate MakePeriodicPayment(final String receiver, final int amount, final int withdrawingWindow, final int period, final int fee, final int timeout, final byte[] lease) throws NoSuchAlgorithmException {
-        if (!Lease.valid(lease)) {
-            throw new IllegalArgumentException("The lease should be an empty array or a 32 byte array.");
-        }
-
+    protected static ContractTemplate MakePeriodicPayment(final String receiver, final int amount, final int withdrawingWindow, final int period, final int fee, final int timeout, final Lease lease) throws NoSuchAlgorithmException {
```
algonautshant: nit
Very long line. Consider breaking for the emacs users :-)

src/main/java/com/algorand/algosdk/transaction/Lease.java
```diff
@@ -0,0 +1,70 @@
+package com.algorand.algosdk.transaction;
+
+import com.algorand.algosdk.util.Encoder;
+import com.fasterxml.jackson.annotation.*;
+
+import java.util.Arrays;
+import java.util.Random;
+
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+public class Lease {
```
algonautshant: Please describe what this class is all about. Also helps adding references to the algorand doc where the concept is presented.

PR #84, opened 2020-01-14
DynamicFee Template

winder:
Implementation of the DynamicFee template. This code has a workaround for the inconsistent fee initialization in order for the golden to match those used by the other SDKs. A followup PR will be made to fix the fee initialization issue because those changes will break many of the existing unit tests.


winder:
@EvanJRichard thanks, I think that was a blocking issue -- fixed (and now my "could you create two separate goldens" question no longer makes sense)

PR #87, opened 2020-01-16
Merge offset and parameter

winder:
More small adjustments to this code to merge the parallel value/offset arrays


src/main/java/com/algorand/algosdk/templates/LimitOrder.java
```diff
@@ -0,0 +1,4 @@
+package com.algorand.algosdk.templates;
+
+public class LimitOrder {
```
winder: Ignore this for now :)

PR #88, opened 2020-01-16
LimitOrder Template

winder:


src/main/java/com/algorand/algosdk/templates/LimitOrder.java
```diff
@@ -1,4 +1,143 @@
...
+        Address owner = new Address(data.byteBlock.get(0));
+        int assetId = data.intBlock.get(6);
+        int ratd = data.intBlock.get(7);
+        int ratn = data.intBlock.get(8);
+
+        // Verify the exchange rate ratio
+        if (assetAmount * ratd < algoAmount * ratn) {
+            throw new IllegalArgumentException("The provided amounts to not meet the contract exchange rate of assetAmount = algoAmount * " + ratn + " / " + ratd);
+        }
+
```
mjiang102628: It might also be good to check for algoAmount > minTrade here since we read the contract anyway.

src/main/java/com/algorand/algosdk/templates/LimitOrder.java
```diff
@@ -1,4 +1,143 @@
...
+        Transaction tx1 = new Transaction(
+                owner,
+                Account.MIN_TX_FEE_UALGOS,
+                BigInteger.valueOf(firstValid),
+                BigInteger.valueOf(lastValid),
+                null,
+                "",
+                genesisHash,
+                BigInteger.valueOf(assetAmount),
```
mjiang102628: Should this be algoAmount?

src/main/java/com/algorand/algosdk/templates/ContractTemplate.java
```diff
@@ -166,4 +166,18 @@ protected static ContractTemplate inject(byte [] program, List<ParameterValue> v
 
 		return new ContractTemplate(updatedProgramByteArray);
 	}
+
+	protected static Logic.ProgramData readAndVerifyContract(byte[] program, int numInts, int numByteArrays) {
+		try {
+			Logic.ProgramData data = Logic.readProgram(program, null);
+
+			if (!data.good || data.intBlock.size() != numInts || data.byteBlock.size() != numByteArrays) {
+				throw new IllegalArgumentException("Invalid contract detected.");
```
EvanJRichard: does this immediately get caught by the surrounding `try` block and changed to the exception on line 180?

src/main/java/com/algorand/algosdk/templates/ContractTemplate.java
```diff
@@ -166,4 +166,18 @@ protected static ContractTemplate inject(byte [] program, List<ParameterValue> v
...
+			}
+
+			return data;
+		} catch (Exception e) {
+			throw new IllegalArgumentException("Unable to parse contract.");
```
EvanJRichard: Is ignoring the thrown exception and throwing a new one the desired behavior, as opposed to letting the thrown exception bubble up? ("Yes" is a perfectly ok answer :+1:)

winder: Oops, no I wanted the different messages.

PR #91, opened 2020-01-24
Remove duplicate 'jsonSerializeDeserializeCheck' methods

winder:
Replace duplicate json encoding test methods with a single method which also tests the msgpack encoding.

Fix a bug that was uncovered by these tests.


src/test/java/com/algorand/algosdk/transaction/TestTransaction.java
```diff
@@ -503,7 +502,7 @@ public void testMakeAssetAcceptanceTxn() throws Exception {
         SignedTransaction stxDecoded = Encoder.decodeFromMsgPack(encodedOutBytes, SignedTransaction.class);
```
algonautshant: Is this check here and elsewhere now redundant?
Isn't this already checked in serializeDeserializeCheck?

winder: I think so, but assuming the golden is shared across SDKs I think it's still a good safeguard.

PR #92, opened 2020-01-27
Use JUnit5.

winder:
This required some build magic to make tests using Java 8, which is required by JUnit5. This has the added benefit of allowing us to use more expressive assertions for exception validation.

It seems to work alright with IntelliJ, but the IDE will think that lambda's are OK, which definitely isn't the case. Anything outside the IDE will properly set the build target, and will fail to compile Java 8 features (assuming they aren't cached, so `mvn clean package` may need to be done).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,8 +33,13 @@
     </developers>
 
     <properties>
-        <maven.compiler.source>1.7</maven.compiler.source>
-        <maven.compiler.target>1.7</maven.compiler.target>
+        <junit5.version>5.5.2</junit5.version>
+        <maven.compiler.plugin.version>3.8.1</maven.compiler.plugin.version>
+
+        <!-- Test sources at 1.8 to allow using JUnit5 -->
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -33,8 +33,13 @@
     </developers>
 
     <properties>
-        <maven.compiler.source>1.7</maven.compiler.source>
-        <maven.compiler.target>1.7</maven.compiler.target>
+        <junit5.version>5.5.2</junit5.version>
+        <maven.compiler.plugin.version>3.8.1</maven.compiler.plugin.version>
+
+        <!-- Test sources at 1.8 to allow using JUnit5 -->
+        <java.version>1.7</java.version>
```
egieseke: Can we update the java version to a more recent version?

winder: No, we need to support Android

winder:
@algonautshant great idea, documentation has been updated.

PR #97, opened 2020-01-29
Add missing asset holding getters.

### LINKED ISSUE #63: [java-sdk] add getter for all properties in public class AssetHolding
Creator and Frozen are private: 
https://github.com/algorand/java-algorand-sdk/blob/master/src/main/java/com/algorand/algosdk/algod/client/model/AssetHolding.java

winder:
The go-algorand swagger generation is broken, so this class was not properly generated. See https://github.com/algorand/go-algorand/issues/739

Added getters and documentation.

This resolves #63

### DIFF
--- a/src/main/java/com/algorand/algosdk/algod/client/model/AssetHolding.java
+++ b/src/main/java/com/algorand/algosdk/algod/client/model/AssetHolding.java
@@ -12,6 +12,7 @@ package com.algorand.algosdk.algod.client.model;
 import com.google.gson.annotations.SerializedName;
 
 import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
 
 @ApiModel(description = "AssetParams specifies the holdings of a particular asset.")
 public class AssetHolding {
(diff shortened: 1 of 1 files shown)


src/main/java/com/algorand/algosdk/algod/client/model/AssetHolding.java
```diff
@@ -24,7 +24,31 @@
   @SerializedName("frozen")
   private Boolean frozen = null;
   
+
+  /**
+   * Creator specifies the address that created this asset. This is the address where the parameters for this asset can be found, and also the address where unwanted asset units can be sent in the worst case.
```
algonautshant: Please break the line  (>200 chars) for those stuck with emacs :-)

src/main/java/com/algorand/algosdk/algod/client/model/AssetHolding.java
```diff
@@ -24,7 +24,31 @@
...
+  @ApiModelProperty(value = "Amount specifies the number of units held.")
   public java.math.BigInteger getAmount() {
       return amount;
   }
+
+  /**
+   * Frozen specifies whether this holding is frozen.
+   * @return frozen
+   */
+  @ApiModelProperty(value = "Frozen specifies whether this holding is frozen.")
```
algonautshant: Will it be possible/helpful to add links? 
Here, the term frozen is not explained. Adding a link to where the term frozen is described can be useful.

PR #98, opened 2020-02-03
Fix/finish split template.

winder:


winder:
@EvanJRichard I'm not sure. I was thinking it could be similar to the LimitOrder situation, and the user should verify their inputs to avoid any surprises.

EvanJRichard:
There's a test error, `[ERROR] /home/travis/build/algorand/java-algorand-sdk/src/test/java/com/algorand/algosdk/templates/TestTemplates.java:[36,49] incompatible types: com.algorand.algosdk.crypto.Address cannot be converted to java.lang.String` but looks like the goldens match the goldens I got :+1:

PR #99, opened 2020-02-12
Template Bug Fixes / Cucumber Test Changes

winder:


src/test/java/com/algorand/algosdk/logic/TestLogic.java
```diff
@@ -162,7 +162,7 @@ public void testCheckProgramInvalidOpcode() throws Exception {
 
         assertThatThrownBy(() -> checkProgram(program, args))
                 .isInstanceOf(IllegalArgumentException.class)
-                .hasMessage("invalid instruction");
+                .hasMessage("invalid instruction: 129");
```
mjiang102628: What does 129 mean here?

winder: That just happens to be the invalid instruction that the program tests:
```
        byte[] program = {
            0x01, 0x20, 0x01, 0x01, (byte)0x81
        };
```
0x81 == 129

src/main/java/com/algorand/algosdk/templates/DynamicFee.java
```diff
@@ -152,13 +152,13 @@ public static SignedDynamicFee SignDynamicFee(final ContractTemplate contract, f
         txn2.setLease(new Lease(txn.lease));
         Account.setFeeByFeePerByte(txn2, BigInteger.valueOf(feePerByte));
 
-        TxGroup.assignGroupID(txn, txn2);
+        TxGroup.assignGroupID(txn2, txn);
```
EvanJRichard: It might help readability to name the `txn`s here, like `mainPay` and `feeReimburse` or `mainTxn` and `feeTxn` or something, rather than numbered txns.

PR #103, opened 2020-03-11
move cucumber tests into java sdk

mjiang102628:
- the original unit tests have been removed for now; junit5 conflicts with cucumber, which uses junit4, and when we add feature files for goldens, the unit tests will have been replaced anyway
- run_docker.sh currently grabs files from the michelle/test branch in algorand-sdk-testing; this will change such that we have a new branch/new directory for each new milestone so that sdks that have not implemented some new feature don't fail because the feature files contain scenarios that it can't complete (for example we might have a branch for before templates, and one that includes templates)


winder:
Deprecated by #106

PR #104, opened 2020-03-12
Make sure note and lease can't be set to an empty byte array.

winder:
It was possible to put the note/lease fields into an invalid state, added some additional checks.

### DIFF
--- a/src/main/java/com/algorand/algosdk/transaction/Transaction.java
+++ b/src/main/java/com/algorand/algosdk/transaction/Transaction.java
@@ -166,7 +166,7 @@ public class Transaction implements Serializable {
         setFee(fee);
         if (firstValid != null) this.firstValid = firstValid;
         if (lastValid != null) this.lastValid = lastValid;
-        if (note != null) this.note = note;
+        setNote(note);
         if (genesisID != null) this.genesisID = genesisID;
         if (genesisHash != null) this.genesisHash = genesisHash;
         if (amount != null) this.amount = amount;
@@ -197,7 +197,7 @@ public class Transaction implements Serializable {
         setFee(fee);
         if (firstValid != null) this.firstValid = firstValid;
         if (lastValid != null) this.lastValid = lastValid;
-        if (note != null) this.note = note;
+        setNote(note);
         if (genesisID != null) this.genesisID = genesisID;
         if (genesisHash != null) this.genesisHash = genesisHash;        
         if (votePK != null) this.votePK = votePK;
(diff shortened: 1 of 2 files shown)


src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -745,6 +745,12 @@ public static Transaction createAssetTransferTransaction(// AssetTransaction
         return tx;
     }
 
+    private void setNote(byte[] note) {
+        if (note != null && note.length != 0) {
```
algonautshant: What is a valid empty note field? How can one reset the note field to empty?

src/test/java/com/algorand/algosdk/transaction/TestTransaction.java
```diff
@@ -655,4 +655,26 @@ public void testTransactionWithLease() throws Exception {
         assertThat(encodedOutBytes).isEqualTo(goldenString);
         TestUtil.serializeDeserializeCheck(stx);
     }
+
+    @Test
+    public void EmptyByteArraysShouldBeRejected() throws Exception {
```
egieseke: Not sure that the semantics of empty string and null byte array are the same.

rotemh:
I believe we should allow these two to be empty. We just need to make sure they get serialize properly.

src/main/java/com/algorand/algosdk/transaction/Transaction.java
winder: In order for the note field to be omitted, it needs to be `null` in the Transaction object. We were previously checking for null, but allowing it to be an array of length zero.

winder:
@rotemh it has always been valid to have them be empty, the problem was that according to the serialization algorithm "empty" means equal to the default value, which is set to null. So if someone set it to an array of size 0, it would not be considered empty.

src/test/java/com/algorand/algosdk/transaction/TestTransaction.java
winder: These fields don't accept a string, so "empty string" doesn't apply to the underlying problem.

src/main/java/com/algorand/algosdk/transaction/Transaction.java
winder: You shouldn't reset the note field, which is why I made this method private. Unfortunately all of the fields are public so it isn't strictly immutable.

Other than the groupID and the fee, I would prefer that the transaction object be immutable.

rotemh:
so in order to not set the `note` or `lease` field one simply should use the `setNote` `setLease`?

winder:
For lease, yes simply skip it. For note, all of the constructors have a note field, you would set it to null (or `new note[]{}`).

PR #106, opened 2020-03-22
Move Cucumber Tests to SDK

winder:
Move stepfiles from the old feature repo into the unit tests.

Update build script to support cucumber + junit5, and disable the integration tests by default.

Add driver script to clone the cucumber feature repo and run the integration tests in a docker container.


pom.xml
```diff
@@ -158,7 +199,38 @@
         <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-surefire-plugin</artifactId>
-            <version>3.0.0-M4</version>
+            <version>${maven.surefire.version}</version>
+            <configuration>
+              <skipTests>true</skipTests>
+            </configuration>
+            <executions>
```
winder: unit tests are always run, integration tests are disabled by default. Enable them with `mvn test -Dskip.integration.tests=false`

pom.xml
```diff
@@ -123,29 +128,65 @@
...
         </dependency>
+
         <dependency>
             <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-engine</artifactId>
+            <artifactId>junit-jupiter-api</artifactId>
             <version>${junit5.version}</version>
             <scope>test</scope>
         </dependency>
+
```
winder: There is a JUnit5 cucumber engine, but it isn't ready for production use. Enabling `junit-vintage-engine` lets us continue to use the JUnit4 cucumber tooling.

PR #110, opened 2020-04-14
Update Java SDK for indexer integration testing.

winder:
Support for the new cucumber with indexer environment.

Changes:
1. Connection information is now static, update SDK to use hard coded token + ports.
2. Docker container only needs to build + run the SDK, so it is greatly simplified (and can leverage an off the shelf container if one exists).
3. The driver script stays mostly the same, but now calls `./scripts/up.sh` to start the test environment, and launches its own container with `--network host` so that it can use `localhost:<port>` in the tests.

### DIFF
--- a/run_integration_tests.sh
+++ b/run_integration_tests.sh
@@ -5,20 +5,22 @@ set -e
 rootdir=`dirname $0`
 pushd $rootdir
 
-# get feature files from algorand-sdk-testing repo
-sudo rm -rf temp
-git clone --single-branch --branch templates https://github.com/algorand/algorand-sdk-testing.git temp
+# Reset test harness
+rm -rf test-harness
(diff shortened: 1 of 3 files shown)


run_integration_tests.sh
```diff
@@ -5,20 +5,22 @@ set -e
 rootdir=`dirname $0`
 pushd $rootdir
 
-# get feature files from algorand-sdk-testing repo
-sudo rm -rf temp
-git clone --single-branch --branch templates https://github.com/algorand/algorand-sdk-testing.git temp
+# Reset test harness
+rm -rf test-harness
+git clone --single-branch --branch will/indexer https://github.com/algorand/algorand-sdk-testing.git test-harness
```
EvanJRichard, line 10: this will need to become branch `develop` if the corresponding sdk-testing PR gets merged yeah?

PR #124, opened 2020-05-05
Update TEAL langspec.json for TEAL v2

pzbitskiy:


winder:
This looks good to me, but we need to wait until after indexer is released before we can merge it in.

PR #129, opened 2020-05-07
Move files around and update integration test runner script.

winder:


algonautshant:
Looks good. 
Thanks for doing this.

PR #132, opened 2020-05-09
Updates to generator refactoring, catchpoint and indexer support

algonautshant:
* Proposed changes to Will's review: generator refactoring #130
 ** Generator spec file path as optional parameter.
 ** Small fixes

* Support for Indexer catchpoint http methods.
* Updates from the new spec file.


winder:
LGTM. But I think that we might be excluding the "private" endpoints from the SDKs, the node status definitely still needed to be updated though.

generator/src/main/java/com/algorand/sdkutils/RunQueryMapperGenerator.java
```diff
@@ -35,7 +35,7 @@ public static void main (String args[]) throws JsonProcessingException, IOExcept
 
 		
 		QueryMapperGenerator qmg = new QueryMapperGenerator(indexerRoot, algodRoot);
-		qmg.writeQueryMapper("src/test/java/com/algorand/algosdk/unit/utils");
+		qmg.writeQueryMapper("../src/test/java/com/algorand/algosdk/unit/utils");
```
winder, line 38: I was thinking that these would go away with the CLI mode. Left them here mainly for example and convenience.

winder:
Are there missing files? I thought I commented out several QueryMapper components

winder:
I'm merging this in, but we may need a followup to remove the catchup endpoint (and participation keys/shutdown endpoints from earlier)

generator/src/main/java/com/algorand/sdkutils/RunQueryMapperGenerator.java
algonautshant: I am definitely in favor or leaving the hard-coded "default" paths.

PR #135, opened 2020-05-10
Round of fixes for algodV2

algonautshant:
* Stopped exposing format query parameter.
* BlockCertificate and BlockHeader are of type HashMap.
* Fixed the in-body required parameter check.


src/main/java/com/algorand/algosdk/v2/client/algod/GetPendingTransactionsByAddress.java
```diff
@@ -28,14 +27,6 @@ public GetPendingTransactionsByAddress(Client client, Address address) {
 		this.address = address;
 	}
 
-	/**
-	 * Configures whether the response object is JSON or MessagePack encoded. 
-	 */
-	public GetPendingTransactionsByAddress format(Enums.Format format) {
-		addQuery("format", String.valueOf(format));
```
winder, line 35: We need to set this somewhere:
addQuery("format", String.valueOf(format));

algonautshant: This is an optional parameter. What does it mean? What is the default value?

PR #141, opened 2020-05-16
JAVA SDK rekeying feature.

algonautshant:
1. In Account class signTransaction and signMultisigTransaction
   - Set authAddr to the signature account if different from the sender account
2. Add rekey method to TransactionBuilder
3. Add publicKey field to SignTransactionRequest
4. Add authAddr to SignedTransaction
5. Add rekeyTo field to Transaction

### DIFF
--- a/src/main/java/com/algorand/algosdk/account/Account.java
+++ b/src/main/java/com/algorand/algosdk/account/Account.java
@@ -136,7 +136,11 @@ public class Account {
         try {
             byte[] prefixEncodedTx = tx.bytesToSign();
             Signature txSig = rawSignBytes(Arrays.copyOf(prefixEncodedTx, prefixEncodedTx.length));
-            return new SignedTransaction(tx, txSig, tx.txID());
+            SignedTransaction stx = new SignedTransaction(tx, txSig, tx.txID());
+            if (!tx.sender.equals(this.address)) {
+            	stx.authAddr(this.address);	
+            }
+            return stx;
         } catch (IOException e) {
             throw new RuntimeException("unexpected behavior", e);
         }
@@ -325,7 +329,11 @@ public class Account {
                 mSig.subsigs.add(new MultisigSubsig(from.publicKeys.get(i)));
             }
         }
-        return new SignedTransaction(tx, mSig, txSig.transactionID);
+        SignedTransaction stx = new SignedTransaction(tx, mSig, txSig.transactionID);
+        if (!tx.sender.equals(this.address)) {
+        	stx.authAddr(this.address);	
+        }
+        return stx;
     }
 
     /**
--- a/src/main/java/com/algorand/algosdk/account/Account.java
+++ b/src/main/java/com/algorand/algosdk/account/Account.java
@@ -138,7 +138,7 @@ public class Account {
             Signature txSig = rawSignBytes(Arrays.copyOf(prefixEncodedTx, prefixEncodedTx.length));
             SignedTransaction stx = new SignedTransaction(tx, txSig, tx.txID());
             if (!tx.sender.equals(this.address)) {
-            stx.authAddr(this.address);
+                stx.authAddr(this.address);
             }
             return stx;
         } catch (IOException e) {
--- a/src/main/java/com/algorand/algosdk/account/Account.java
+++ b/src/main/java/com/algorand/algosdk/account/Account.java
@@ -259,12 +259,17 @@ public class Account {
      * @return an estimated byte size for the transaction.
      */
     public static BigInteger estimatedEncodedSize(Transaction tx) throws NoSuchAlgorithmException {
-        Account acc = new Account();
-        try {
-            return BigInteger.valueOf(Encoder.encodeToMsgPack(acc.signTransaction(tx)).length);
-        } catch (IOException e) {
-            throw new RuntimeException("unexpected behavior", e);
-        }
+    	try {
+    		long length = Encoder.encodeToMsgPack(
+    				new SignedTransaction(
+    						tx, 
+    						new Account().rawSignBytes(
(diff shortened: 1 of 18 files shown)


src/main/java/com/algorand/algosdk/transaction/SignedTransaction.java
```diff
@@ -65,6 +68,11 @@ public SignedTransaction(Transaction tx, LogicsigSignature lSig, String txId) {
     private SignedTransaction() {
     }
 
+    public SignedTransaction authAddr(Address authAddr) {
```
winder, line 71: Is there a way to make this private? At the very minimum we'll need a javadoc comment here stating that it should not be set manually.

src/test/java/com/algorand/algosdk/unit/Rekeying.java
```diff
@@ -0,0 +1,75 @@
...
+import com.algorand.algosdk.account.Account;
+import com.algorand.algosdk.builder.transaction.PaymentTransactionBuilder;
+import com.algorand.algosdk.builder.transaction.TransactionBuilder;
+import com.algorand.algosdk.transaction.SignedTransaction;
+
+import io.cucumber.java.en.Given;
+import io.cucumber.java.en.Then;
+import io.cucumber.java.en.When;
+
+public class Rekeying {
```
winder, line 15: If you include `Stepdefs` in this file, do you still need to re-implement these?
```
src/test/java/com/algorand/algosdk/integration/Stepdefs.java
243:    @Given("payment transaction parameters {int} {int} {int} {string} {string} {string} {int} {string} {string}")

src/test/java/com/algorand/algosdk/integration/Stepdefs.java
281:    @Given("mnemonic for private key {string}")

src/test/java/com/algorand/algosdk/integration/Stepdefs.java
900:    @When("I create the flat fee payment transaction")

src/test/java/com/algorand/algosdk/integration/Stepdefs.java
353:    @When("I sign the transaction with the private key")

src/test/java/com/algorand/algosdk/integration/Stepdefs.java
358:    @Then("the signed transaction should equal the golden {string}")
```

src/main/java/com/algorand/algosdk/account/Account.java
```diff
@@ -255,12 +259,17 @@ static public void setFeeByFeePerByte(Transaction tx, BigInteger suggestedFeePer
...
+    		long length = Encoder.encodeToMsgPack(
+    				new SignedTransaction(
+    						tx, 
+    						new Account().rawSignBytes(
+    								Arrays.copyOf(tx.bytesToSign(), tx.bytesToSign().length)),
+    						tx.txID())).length;
+    		return BigInteger.valueOf(length);
+    	} catch (IOException e) {
+    		throw new RuntimeException("unexpected behavior", e);
+    	}
```
winder, line 272: mixed tabs/spaces here

winder: Could you explain what is happening here? I suppose this is to account for the AuthAddr being set in the signed transaction?

Is this to match with the golden, and do the legacy tests need to be updated?

Was something like this needed in the other SDKs?

algonautshant:
> I'm curious to hear your thoughts about the `Account` design. There seem to be some issues here. On one hand the metaphor of having an account that signs a transaction makes sense. But all of the other SDKs have signing as an action done on a Transaction. In this case there is a design problem where the Account must know about the internal `AuthAddr` field.

If I understand you correctly, I agree with you. The account depends on transaction (to sign it), and the transaction depends on the account.

I once refactored the code, but there was pushback. 
You can see the discussion here: https://github.com/algorand/java-algorand-sdk/pull/36

src/main/java/com/algorand/algosdk/transaction/SignedTransaction.java
algonautshant: No, Account calls it while signing. I'll add a comment.

src/test/java/com/algorand/algosdk/unit/Rekeying.java
algonautshant: Maybe there is a way, but cucumber was looking for those implementations here.

src/main/java/com/algorand/algosdk/account/Account.java
algonautshant: The transaction is signed with a random account to get the size of the signed transaction. SignTransaction does the signing, but also sets authAddr which is not desired here.
Legacy tests do not need to be updated. 
This is the consequence of having the Account sign a Transaction, and being aware of authAddr.

winder: Ok, if the existing tests still pass that should be enough. I think our program size estimates aren't perfect in general but that is outside the scope of this feature.

src/main/java/com/algorand/algosdk/builder/transaction/AssetConfigureTransactionBuilder.java
```diff
@@ -39,7 +42,7 @@
     }
 
     @Override
-    protected Transaction buildInternal() {
+    protected void buildInternal() {
```
winder, line 45: I think this is OK because buildInternal is not public, but we should avoid adding the txn field. One way to do this would be by repurposing this method from "build" to "apply". Perhaps `protected void applyTo(Transaction txn)`.

Then build would be something like:
```
build() {
    Transaction txn = new Transaction();
    this.applyTo(txn);
}
```

No need to do that now though, it looks like the behavior here is correct, and I'm really happy to see those deprecated function calls go away.

algonautshant: I like the idea. I think I just go ahead and make the change. 
Thanks.

PR #151, opened 2020-06-03
Application Transaction Support

winder:
The application integration test currently depends on running the test environment with `up.sh --type source`, and setting the source branch to the applications branch. So travis is currently failing.

Once applications are merged in I'll refresh the tests and everything should pass.


src/main/java/com/algorand/algosdk/crypto/TEALProgram.java
```diff
@@ -0,0 +1,48 @@
+package com.algorand.algosdk.crypto;
+
+import com.algorand.algosdk.logic.Logic;
+import com.algorand.algosdk.util.Encoder;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+import java.io.IOException;
```
algonautshant, line 9: IOException is not used.

src/main/java/com/algorand/algosdk/logic/StateSchema.java
```diff
@@ -0,0 +1,48 @@
+package com.algorand.algosdk.logic;
+
+import com.algorand.algosdk.crypto.Address;
```
algonautshant, line 3: Address not used.

src/main/java/com/algorand/algosdk/logic/StateSchema.java
```diff
@@ -0,0 +1,48 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+import java.io.Serializable;
+import java.math.BigInteger;
+import java.util.Arrays;
```
algonautshant, line 10: Arrays not used.

src/main/java/com/algorand/algosdk/transaction/AssetParams.java
```diff
@@ -0,0 +1,161 @@
...
+                        @JsonProperty("f") byte[] assetFreeze,
+                        @JsonProperty("c") byte[] assetClawback) {
+        this(assetTotal, assetDecimals, assetDefaultFrozen, assetUnitName, assetName, url, metadataHash, new Address(assetManager), new Address(assetReserve), new Address(assetFreeze), new Address(assetClawback));
+    }
+
+    /**
+     * Convert the given object to string with each line indented by 4 spaces
+     * (except the first line).
+     */
+    private String toIndentedString(Object o) {
```
algonautshant, line 155: Looks like this method is no longer used.

src/test/java/com/algorand/algosdk/integration/Applications.java
```diff
@@ -0,0 +1,185 @@
+package com.algorand.algosdk.integration;
+
+import com.algorand.algosdk.account.Account;
+import com.algorand.algosdk.builder.transaction.*;
+import com.algorand.algosdk.crypto.Address;
+import com.algorand.algosdk.crypto.TEALProgram;
```
algonautshant, line 6: not used.

src/test/java/com/algorand/algosdk/integration/Applications.java
```diff
@@ -0,0 +1,185 @@
...
+import com.algorand.algosdk.logic.StateSchema;
+import com.algorand.algosdk.transaction.SignedTransaction;
+import com.algorand.algosdk.transaction.Transaction;
+import com.algorand.algosdk.util.ConversionUtils;
```
algonautshant, line 10: not used.

src/test/java/com/algorand/algosdk/integration/Applications.java
```diff
@@ -0,0 +1,185 @@
...
+import com.algorand.algosdk.util.Encoder;
+import com.algorand.algosdk.util.ResourceUtils;
```
algonautshant, line 12: not used.

src/test/java/com/algorand/algosdk/integration/Applications.java
```diff
@@ -0,0 +1,185 @@
...
+import com.algorand.algosdk.v2.client.common.AlgodClient;
```
algonautshant, line 13: not used.

src/test/java/com/algorand/algosdk/integration/Applications.java
```diff
@@ -0,0 +1,185 @@
...
+import com.algorand.algosdk.v2.client.common.Response;
+import com.algorand.algosdk.v2.client.model.PendingTransactionResponse;
+import com.algorand.algosdk.v2.client.model.PostTransactionsResponse;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.cucumber.java.en.Given;
+import org.apache.commons.lang3.StringUtils;
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
```
algonautshant, line 21: not used.

src/test/java/com/algorand/algosdk/unit/Base.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class Base {
+    private Map<String, Account> signingAccounts = new HashMap<>();
+
+    /**
+     * Helper to sign a transaction with the correct account, or fail the test if the correct account isn't available.
+     */
+    public SignedTransaction signTransaction(Transaction tx) throws NoSuchAlgorithmException {
+        String signingAddress = tx.sender.toString();
+        if (tx.rekeyTo != null && !tx.rekeyTo.equals(new Address())) {
```
algonautshant, line 23: Maybe we should add a getSigerAddress method?

src/main/java/com/algorand/algosdk/builder/transaction/ApplicationCreateTransactionBuilder.java
```diff
@@ -0,0 +1,74 @@
...
+        txn.globalStateSchema = globalStateSchema;
+
+        super.applyTo(txn);
+    }
+
+    /**
+     * This option is disabled for application create, where the ID must be changed from 0.
+     */
+    @Override
+    public T applicationId(Long appId) {
```
algonautshant, line 35: Why do we need this method? Better to do away with it.

algonautshant: I think I understand now.
It can also go to all the leaf classes, and not the base class, and avoid it here. 
I think it is okay as is.

winder: Yes thats right. The supported fields hierarchy didn't work out quite right, so I either "disable" the method here or have an extra base-class containing everything except this method.

PR #157, opened 2020-06-13
Generator enhancements and go-sdk generator 

algonautshant:
In this change, the generator is enhanced to be a general purpose
generator.

Publisher/Subscriber paradigm is added to build generators outside the
spec file parser.

Relevant events such as new model or new parameter, with the relevant
customizations, e.g. type conversion are published, and subscribers
can pick up the messages and generate code.

The first use case of the publisher/subscriber is implemented for
generating go-sdk code.

### DIFF
--- a/generator/src/main/java/com/algorand/sdkutils/RunIndexerGenerator.java
+++ b/generator/src/main/java/com/algorand/sdkutils/RunIndexerGenerator.java
@@ -21,6 +21,7 @@ public class RunIndexerGenerator {
                 "../src/main/java/com/algorand/algosdk/v2/client/common",
                 "com.algorand.algosdk.v2.client.common",
                 "X-Indexer-API-Token",
-                true);
+                true,
+                "./go-sdk");
     }
 }
--- a/generator/src/main/java/com/algorand/sdkutils/generators/Generator.java
+++ b/generator/src/main/java/com/algorand/sdkutils/generators/Generator.java
@@ -44,7 +44,7 @@ public class Generator {
 
     // Get TypeDef for Address type. 
     // It provides the special getter/setter needed for this type.
-    static TypeDef getAddress(String propName,
+    static TypeDef getAddress(String propName, String goPropertyName,
             Map<String, Set<String>> imports, boolean forModel, String desc, boolean required) {
         addImport(imports, "com.algorand.algosdk.crypto.Address");
         if (forModel) {
@@ -67,12 +67,13 @@ public class Generator {
                 "        }\n" +
                 "    }\n" + 
                 "    public Address " + javaName + ";\n");
-        return new TypeDef("Address", "address", sb.toString(), "getterSetter", propName, desc, required);
+        return new TypeDef("Address", "address", sb.toString(), "getterSetter", propName, 
+                goPropertyName, desc, required);
     }
 
     // Get base64 encoded byte[] type.
     // It provides the special getter/setter needed for this type
-    static TypeDef getBase64Encoded(String propName, String rawType,
+    static TypeDef getBase64Encoded(String propName, String goPropertyName, String rawType,
             Map<String, Set<String>> imports, boolean forModel, String desc, boolean required) {
         if (imports != null) {
             addImport(imports, "com.algorand.algosdk.util.Encoder");
@@ -89,12 +90,13 @@ public class Generator {
                 "    }\n" +
                 "    public byte[] "+ javaName +";\n");
         // getterSetter typeName is only used in path.
-        return new TypeDef("byte[]", "byteArray", sb.toString(), "getterSetter", propName, desc, required);
+        return new TypeDef("byte[]", "byteArray", sb.toString(), "getterSetter", 
+                propName, goPropertyName, desc, required);
     }
 
     // Get array type of base64 encoded bytes.
     // It provides the special getter/setter needed for this type
-    static TypeDef getBase64EncodedArray(String propName, String rawType,
+    static TypeDef getBase64EncodedArray(String propName, String goPropertyName, String rawType,
             Map<String, Set<String>> imports, boolean forModel, String desc, boolean required) {
         if (forModel == false) {
             throw new RuntimeException("array of byte[] cannot yet be used in a path or path query.");
@@ -123,12 +125,13 @@ public class Generator {
                 "     }\n" +
                 "    public List<byte[]> " + javaName + ";\n");
         // getterSetter typeName is only used in path.
-        return new TypeDef("", rawType, sb.toString(), "getterSetter,array", propName, desc, required);
+        return new TypeDef("", rawType, sb.toString(), "getterSetter,array", 
+                propName, goPropertyName, desc, required);
     }
 
     // Get array type of base64 encoded bytes.
     // It provides the special getter/setter needed for this type
-    static TypeDef getEnum(JsonNode prop, String propName) {
+    static TypeDef getEnum(JsonNode prop, String propName, String goPropertyName) {
         JsonNode enumNode = prop.get("enum");
         if (enumNode == null) {
             throw new RuntimeException("Cannot find enum info in node: " + prop.toString());
@@ -163,7 +166,7 @@ public class Generator {
         enumClassName = "Enums." + enumClassName;
         String desc = prop.get("description") == null ? "" : prop.get("description").asText();
         return new TypeDef(enumClassName, prop.get("type").asText(), 
-                sb.toString(), "enum", propName, desc, isRequired(prop));
+                sb.toString(), "enum", propName, goPropertyName, desc, isRequired(prop));
     }
 
     // getType returns the type fron the JsonNode
@@ -172,7 +175,9 @@ public class Generator {
             boolean asObject,
             Map<String, Set<String>> imports,
             String propName, boolean forModel) {        
-        String desc = prop.get("description") == null ? "" : prop.get("description").asText();  
+        String desc = prop.get("description") == null ? "" : prop.get("description").asText();
+        String goName = prop.get("x-go-name") != null ? 
+                prop.get("x-go-name").asText() : "";
         if (prop.get("$ref") != null) {
             JsonNode typeNode = prop.get("$ref");
             String type = getTypeNameFromRef(typeNode.asText());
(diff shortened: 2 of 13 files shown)


generator/src/main/java/com/algorand/sdkutils/generators/Generator.java
```diff
@@ -215,57 +205,65 @@ static TypeDef getType(
...
-                return new TypeDef("HashMap<String,Object>");
+                return new TypeDef("HashMap<String,Object>", type, propName, goName, desc, isRequired(prop));
             }
         }
         switch (type) {
         case "integer":
-            return asObject ? new TypeDef("Long") : new TypeDef("long");
+            String longName = asObject ? "Long" : "long"; 
+            return new TypeDef(longName, type, propName, goName, desc, isRequired(prop));
         case "object":
```
winder, line 241: I'm not sure it's valid to map object -> string, maybe we don't actually hit this case currently?

generator/src/main/java/com/algorand/sdkutils/generators/Generator.java
```diff
@@ -317,14 +319,16 @@ static String getImports(Map<String, Set<String>> imports) {
...
         StringBuffer buffer = new StringBuffer();
 
         if (typeObj.isOfType("getterSetter")) {
             return typeObj.def.toString();
         }
 
-        switch (typeObj.typeName) {
+        switch (typeObj.javaTypeName) {
         case "java.util.DateTime":
             throw new RuntimeException("Parsing of time is not yet implemented!");
```
winder, line 333: I thought we had a couple date fields?

generator/src/main/java/com/algorand/sdkutils/utils/Tools.java
```diff
@@ -0,0 +1,77 @@
+package com.algorand.sdkutils.utils;
+
+import java.util.StringTokenizer;
+
+public class Tools {
+
+    // formatComment formats the comment by breaking lines, and incorporating 
+    // embedded formatting inside the comment. 
+    // If the comment is embedded inside another comment, full == false
+    public static String formatComment(String comment, String tab, boolean full) {
```
winder, line 10: We should do the HTML encoding, apache-commons has a `StringUtils.encodeHtml` helper that should be able to do this for us.

generator/src/main/java/com/algorand/sdkutils/listeners/Publisher.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+
+import com.algorand.sdkutils.listeners.Publisher.Events;
+import com.algorand.sdkutils.utils.StructDef;
+import com.algorand.sdkutils.utils.TypeDef;
+import com.fasterxml.jackson.databind.JsonNode;
+
+public class Publisher {
```
winder, line 12: I'm a fan of guava's EventBus for a loosely coupled publish/subscribe library: https://github.com/google/guava/wiki/EventBusExplained

No need to change over since you've already written it

generator/src/main/java/com/algorand/sdkutils/generators/Generator.java
```diff
@@ -215,57 +205,65 @@ static TypeDef getType(
...
+    // one is relevant here. 
     public static String getTypeFormat(JsonNode typeNode, String propName) {
         String format = typeNode.get("x-algorand-format") != null ? typeNode.get("x-algorand-format").asText() : "";
         String type = typeNode.get("type").asText();
         format = typeNode.get("format") != null && format.isEmpty() ? typeNode.get("format").asText() : format;
-        format = typeNode.get("x-go-name") != null && format.isEmpty() ? typeNode.get("x-go-name").asText() : format;
-        if ((propName.equals("address") || propName.contentEquals("accountId")) &&
+        if ((propName.equals("address") || 
+                propName.contentEquals("account-id") || 
+                propName.contentEquals("AccountID")) &&
```
winder, line 266: Why are these fields hard coded? should we update the spec?

generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
```diff
@@ -0,0 +1,678 @@
...
+
+    public void terminate() {
+        modelWriter.close();
+        modelWriter = null;
+
+        writeClientFunctions();
+    }
+
+    private void writeClientFunctions() {
+        BufferedWriter bw = newFile("applicationclient", filesFolder);
```
winder, line 102: Should applicationclient be configurable?

winder: Looks like there are several missing configurations here if we want to use this for the indexer and algod client. If we're doing this, we need to be sure to use the generator for both.

generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
```diff
@@ -0,0 +1,678 @@
...
+        case "TransactionKeyreg":
+        case "TransactionPayment":
+        case "TransactionSignature":
+        case "TransactionSignatureLogicsig":
+        case "TransactionSignatureMultisig":
+        case "TransactionSignatureMultisigSubsignature":
+            goType = type;
+            break;
+        default:
+            goType = "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXx";
```
winder, line 518: Is this actually hit? If it's an error it should be an error.

generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
```diff
@@ -0,0 +1,678 @@
...
+import java.util.StringTokenizer;
+import java.util.TreeMap;
+import java.util.TreeSet;
+
+import com.algorand.sdkutils.listeners.Publisher.Events;
+import com.algorand.sdkutils.utils.StructDef;
+import com.algorand.sdkutils.utils.Tools;
+import com.algorand.sdkutils.utils.TypeDef;
+
+public class GoGenerator extends Subscriber {
```
winder, line 21: I really think we need some sort of template system to generate the code. Once we build up all the models/paths/parameters we should be able to pipe the metadata into a series of templates which could be swapped out based on the SDK being generated.

We could either have a property file that maps "types" to SDK types, have a thin per-sdk mapping function, or use the template language to convert them in each template. I prefer keeping the SDK specific business logic in the template engine.

If we don't do something like that it seems like we're going to have 4 nearly identical implementations for this fairly large and complex file.

generator/src/main/java/com/algorand/sdkutils/utils/TypeDef.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+        return this.type.contains(type);
+    }
+
+    @Override
+    public String toString() {
+        throw new RuntimeException("Should not get the string value of the object directly!");
+    }
+
+    public String javaTypeName;
```
winder, line 62: Would this be factored out if we converted the java generator to use the publish/subscribe system?

generator/src/main/java/com/algorand/sdkutils/generators/Generator.java
algonautshant: Definitely helps to update the specs. 
This is around updating the specs.

generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
algonautshant: Yes, you are correct. Next time around we need to generate those from the generator, we should do it. 
This is a shortcut in this PR, since we did not need more for this task.

generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
algonautshant: Yes, this needs to throw an exception. However, it is convenient to know what is missing by just looking into the generated code, without going to the debugger.

generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
algonautshant: Your observation is in place.
However, I worry that "4 nearly identical" may be problematic. I think individual customizations might be more difficult with templates than individual java programs.

generator/src/main/java/com/algorand/sdkutils/utils/Tools.java
algonautshant: This can be future enhancement?

generator/src/main/java/com/algorand/sdkutils/generators/Generator.java
algonautshant: Yes. We have it for paths. We encode dates but not decode them from responses.

generator/src/main/java/com/algorand/sdkutils/generators/Generator.java
algonautshant: e.g. HealthCheck hits this. This will be refined in due time (i.e. testing for applications).

generator/src/main/java/com/algorand/sdkutils/RunIndexerGenerator.java
```diff
@@ -21,6 +21,7 @@ public static void main (String args[]) throws Exception {
                 "../src/main/java/com/algorand/algosdk/v2/client/common",
                 "com.algorand.algosdk.v2.client.common",
                 "X-Indexer-API-Token",
-                true);
+                true,
+                "./go-sdk");
```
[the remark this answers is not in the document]
algonautshant: I agree. This can be enhanced for that purpose.

generator/src/main/java/com/algorand/sdkutils/utils/TypeDef.java
algonautshant: Yes. This and a lot more may need to change for that purpose.
Maybe after the release?

generator/src/main/java/com/algorand/sdkutils/utils/Tools.java
winder: yes

PR #160, opened 2020-06-22
Allow Client 'host' field to act like a URI

winder:
There was no way to specify `https`. Rather than add another parameter to the constructor I upgraded the `host` field to accept URI formatting.


src/main/java/com/algorand/algosdk/v2/client/common/AlgodClient.java
```diff
@@ -17,6 +17,12 @@
 
 public class AlgodClient extends Client {
 
+    /**
+     * Construct a AlgodClient for communicating with the REST API.
+     * @param host host in a plain "localhost" format or a URI format like "https://localhost".
```
algorotem, line 22: I think "plain "localhost" format" is not a thing. It's essentially a URI without the protocol specified.  We should specify that we expect a URI and if the user doesn't contain the scheme (http/https), the client assumes HTTP.

src/main/java/com/algorand/algosdk/v2/client/common/Client.java
```diff
@@ -30,7 +30,21 @@ public String getHost() {
...
+            httpUrlBuilder = parsedHttpUrl.newBuilder();
+
+            // If the default port is being used, override it with the port provided to our constructor.
+            if (HttpUrl.defaultPort(parsedHttpUrl.scheme()) == parsedHttpUrl.port()) {
+                httpUrlBuilder.port(port);
+            }
+        } else {
+            httpUrlBuilder = new HttpUrl.Builder();
+            httpUrlBuilder.scheme("http").port(port).host(host);
+        }
```
algorotem, line 46: I wouldn't correct the user's behavior. If the user provided 2 different ports, throw an error. It's ok to request the port in a separate field.

winder:
Moved to hotfix

PR #164, opened 2020-06-25
Generator Fixes

algonautshant:
* Various generator fixes to enable algod generation for go-algorand-sdk
* Workaround after fixing client method names first letter case issue
  for java:
  The method names for the algod client started with an upper case
  letter.
  This was because the names came from the spec file with upper case
  first letter for algod, but lower case for the indexer. Hence, this
  issue is only in algod v2. When the formatter was fixed to lower the
  case for method names, had to keeps these names upper case not to
  change the published apis.
* Comments of the client constructors (indexer and algod) were added to
  the generated file. Updated the generator to write these comments.
* Formatting fix for comments: eliminte extra spaces.
* Added check for null address before extracting the string encoding. This
  is needed because auth-addr can be null.
* Updated AccountInformation (Account model) to have AuthAddr 
* Rekeying fields in indexer


generator/src/main/java/com/algorand/sdkutils/RunAlgodV2Generator.java
```diff
@@ -21,7 +21,7 @@ public static void main (String args[]) throws Exception {
                 "com.algorand.algosdk.v2.client.common",
                 "X-Algo-API-Token",
                 false, 
-                "");
+                "./go-sdk-algodv2");
```
winder, line 24: Does this need to be added to the CLI?

src/main/java/com/algorand/algosdk/v2/client/algod/PendingTransactionInformation.java
```diff
@@ -9,21 +9,21 @@
...
- * - transaction committed (committed round > 0) - transaction still in the pool 
- * (committed round = 0, pool error = "") - transaction removed from pool due to 
- * error (committed round = 0, pool error != "") 
- * Or the transaction may have happened sufficiently long ago that the node no 
- * longer remembers it, and this will return an error. 
- * /v2/transactions/pending/{txid} 
+ * Given a transaction id of a recently submitted transaction, it returns
+ * information about it. There are several cases when this might succeed:
+ * - transaction committed (committed round > 0) - transaction still in the pool
+ * (committed round = 0, pool error = "") - transaction removed from pool due to
```
winder, line 15: Looks like we're removing the newlines in the description here.

src/main/java/com/algorand/algosdk/v2/client/common/AlgodClient.java
```diff
@@ -27,108 +27,108 @@ public AlgodClient(String host, int port, String token) {
...
-     * information about it. There are several cases when this might succeed: 
-     * - transaction committed (committed round > 0) - transaction still in the pool 
-     * (committed round = 0, pool error = "") - transaction removed from pool due to 
-     * error (committed round = 0, pool error != "") 
-     * Or the transaction may have happened sufficiently long ago that the node no 
-     * longer remembers it, and this will return an error. 
-     * /v2/transactions/pending/{txid} 
+     * Given a transaction id of a recently submitted transaction, it returns
+     * information about it. There are several cases when this might succeed:
+     * - transaction committed (committed round > 0) - transaction still in the pool
```
winder, line 126: missing newlines here also

generator/src/main/java/com/algorand/sdkutils/RunAlgodV2Generator.java
algonautshant: no. I did not know who was using this.

src/main/java/com/algorand/algosdk/v2/client/algod/PendingTransactionInformation.java
algonautshant: Can you be more specific? Which line is removed? 
It is only removing the space at the end of the line.

src/main/java/com/algorand/algosdk/v2/client/common/AlgodClient.java
algonautshant: Is any new line dropped in this PR? Or this is an existing issue?
If this is not caused in this PR, I rather fix it in another PR.

generator/src/main/java/com/algorand/sdkutils/generators/Utils.java
```diff
@@ -159,17 +159,28 @@ public static void generateClientFile(String clientName, Collection<String> impo
         StringBuffer sb = new StringBuffer();
         sb.append("package " + packageName + ";\n\n");
         sb.append(imports);
-        sb.append("\n\n");
+        sb.append("\n\n");
```
algorandskiy, line 162: seems like exceed whitespaces

algonautshant: Fixed!

src/main/java/com/algorand/algosdk/v2/client/algod/PendingTransactionInformation.java
winder: You can see that the comment is malformed, the bulleted list is missing its newlines.

algonautshant: This was like this since a long time ago. I will look into this again. The fix may be in the generator or in the spec file comments. We'll see. 
Also, the comments are not go lint compliant. The names should start with the method name. Will address these later.

PR #165, opened 2020-06-26
Generator refactoring.

winder:
* `Generator` -> `OpenApiParser`
* `legacyMode` flag in `OpenApiParser` so that the parser can run without supplying java configurations.
* `parse()` function in `OpenApiParser` to more intuitively run the operations done by the static `Generate` method (and only the required operations).
* Improved CLI parsing to allow multiple subcommands.
* Stubbed in `responses` subcommand.

### DIFF
--- a/generator/src/main/java/com/algorand/sdkutils/Generate.java
+++ b/generator/src/main/java/com/algorand/sdkutils/Generate.java
@@ -1,0 +1,94 @@
+package com.algorand.sdkutils;
+
+import com.algorand.sdkutils.generators.OpenApiParser;
+import com.algorand.sdkutils.generators.Utils;
+import com.algorand.sdkutils.listeners.GoGenerator;
+import com.algorand.sdkutils.listeners.Publisher;
+import com.fasterxml.jackson.databind.JsonNode;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.nio.file.Files;
+import java.util.Collection;
+
+public class Generate {
+    /**
+     * This is deprecated. Use OpenApiParser:parse() instead.
+     *
+     * @param clientName     Name of the client class. i.e. IndexerClient
+     * @param specfile       Full path to the OpenAPI v2 spec file.
+     * @param modelPath      Path where generated model files will be placed.
+     * @param modelPackage   Package name to put at the top of generated model files.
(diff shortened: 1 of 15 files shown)


generator/src/main/java/com/algorand/sdkutils/Generate.java
```diff
@@ -0,0 +1,94 @@
...
+     * @param modelPackage   Package name to put at the top of generated model files.
+     * @param pathsPath      Path where generated path builder files will be placed.
+     * @param pathsPackage   Package name to put at the top of generated path builder files.
+     * @param commonPath     Path where the common files are, this is where the client class will be placed.
+     * @param commonPackage  Package name to put at the top of generated client class.
+     * @param tokenName      Name of the token used for this application. i.e. X-Algo-API-Token
+     * @param tokenOptional  Whether or not a no-token version of the constructor should be created.
+     * @param goDirectory    When specified, will generate go code.
+     */
+    @Deprecated
```
algonautshant, line 30: We don't need to deprecate. 
If we have the equivalent one, then just delete this, and use OpenApiParser:parse

generator/src/main/java/com/algorand/sdkutils/generators/OpenApiParser.java
```diff
@@ -20,11 +18,40 @@
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.algorand.sdkutils.utils.*;
 
-public class Generator {
+public class OpenApiParser {
     public static final String TAB = "    ";
 
     protected JsonNode root;
     protected Publisher publisher;
+    protected final boolean legacyMode;
```
algonautshant, line 26: let's name it javaGeneration. 'legacy' is not informative and confusing.

generator/src/main/java/com/algorand/sdkutils/Generate.java
winder: It's still needed, there are some extra steps that is only used by the java generator in this one.

The GoGenerator could now migrate pretty easily though.

generator/src/main/java/com/algorand/sdkutils/generators/OpenApiParser.java
winder: Sounds good, I'll rename it in my other PR.

PR #168, opened 2020-07-02
Applications REST Client Updates

winder:
Regenerate the clients and implement the new response unit tests.


src/main/java/com/algorand/algosdk/v2/client/model/TransactionApplication.java
```diff
@@ -0,0 +1,123 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Fields for application transactions.
+ * Definition:
+ * data/transactions/application.go : ApplicationCallTxnFields
+ */
+public class TransactionApplication extends PathResponse {
+
+    // TODO: Generator needs to be updated to setup the Address like this.
```
winder, line 20: This was added by hand for now

src/main/java/com/algorand/algosdk/v2/client/model/DryrunApp.java
```diff
@@ -0,0 +1,35 @@
+package com.algorand.algosdk.v2.client.model;
```
algorandskiy, line 1: this file should not be created, DryrunApp is gone and has no references

src/main/java/com/algorand/algosdk/v2/client/algod/TealDryRun.java
```diff
@@ -0,0 +1,52 @@
+package com.algorand.algosdk.v2.client.algod;
```
algorandskiy, line 1: this file should be deleted - see TealDryrun.java below

src/main/java/com/algorand/algosdk/v2/client/model/DryrunApp.java
winder: Thanks for catching this.

winder: Looks like we're good now. I didn't see anything while grepping around for "DryRun" and "DryrunApp". Hopefully these are the only unused files that snuck in.

PR #171, opened 2020-07-07
Generator refactoring and fixes

algonautshant:
* Removed java code generation from OpenApiParser to JavaGenerator
* GoGenerator fixes
* Enum class generation is the union of multi spec files as long as all are processed in one execution.


generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
```diff
@@ -326,6 +328,7 @@ private StringBuffer spaces (int c) {
         }
```
winder, line 328: Either return a string, or use `StringBuilder`, no need for this helper to decide whether or not the caller needs thread safety

generator/src/main/java/com/algorand/sdkutils/listeners/JavaGenerator.java
```diff
@@ -1,4 +1,933 @@
...
+            String tokenName, 
+            Boolean tokenOptional) {
+
+        if (packagePath.endsWith("/")) {
+            new Exception("Path shouldn't have a trailing slash.").printStackTrace();;
+        }
+
+        importLines.append("import com.algorand.algosdk.crypto.Address;");
+
+        StringBuffer sb = new StringBuffer();
```
winder, line 192: Use `StringBuilder`, no need for thread safety here.

generator/src/main/java/com/algorand/sdkutils/listeners/JavaGenerator.java
```diff
@@ -1,4 +1,933 @@
...
+            }
+            JavaGenerator.append(currentModelBuffer, "public class " + className + " extends PathResponse {\n\n");
+        }
+
+        pendingOpenFile = true;
+        modelPropertyAdded = false;
+    }
+
+    // Returns an iterator in sorted order of the properties (json nodes). 
+    Iterator<Entry<String, TypeDef>> getSortedProperties() {
```
winder, line 762: This helper seems like an unnecessary abstraction. Either change the cache to an `ArrayList` and call `sort()` in here, or get rid of the helper.

generator/src/main/java/com/algorand/sdkutils/listeners/JavaGenerator.java
```diff
@@ -1,4 +1,933 @@
...
+
+    }
+
+    // getPropertyWithJsonSetter formats the property into java declaration type with 
+    // the appropriate json annotation.
+    void getPropertyWithJsonSetter(TypeDef typeObj, StringBuilder buffer) {
+        String jprop = typeObj.propertyName;
+        String javaName = Tools.getCamelCase(jprop, false);
+        if (typeObj.isOfType("getterSetter")) {
+            writeExpandedDefinitions(typeObj, buffer, imports, true);
```
winder, line 817: I still don't understand `getterSetter`. Maybe that should be removed from the parser and handled in the switch below. There could be a case for `Address`, `binary`, `byte`, etc, and each SDK can handle them however it see's fit

generator/src/main/java/com/algorand/sdkutils/listeners/JavaGenerator.java
```diff
@@ -1,4 +1,933 @@
...
+        }
+        
+        switch (typeObj.javaTypeName) {
+        case "java.util.DateTime":
+            throw new RuntimeException("Parsing of time is not yet implemented!");
+        case "String":
+        case "Long":
+        case "long":
+        case "boolean":
+        case "java.math.BigInteger":
```
winder, line 832: Don't use `javaTypeName` here, you can use the regular type instead and delete javaTypeName from the codebase.

For example check `x-algorand-type` for `unit64` instead of `BigInteger`. Also I still believe that we should pass all relevant information to the listeners - `type`, `x-algorand-type` and `format`.

src/main/java/com/algorand/algosdk/v2/client/algod/GetBlock.java
```diff
@@ -24,6 +24,14 @@ public GetBlock(Client client, Long round) {
         this.round = round;
     }
 
+    /**
+     * The round from which to fetch block information.
+     */
+    public GetBlock round(Long round) {
```
winder, line 30: Path parameters shouldn't be added as query parameters.

generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
```diff
@@ -523,7 +526,8 @@ TypeConverter goType(String type, boolean array, boolean asType, String paramNam
         case "DryrunSource":
         case "DryrunTxnResult":
         case "EvalDeltaKeyValue":
-        case "StaleDelta":            
+        case "StaleDelta":
+        case "DryrunRequest":
```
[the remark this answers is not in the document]
algonautshant: I agree.

generator/src/main/java/com/algorand/sdkutils/listeners/JavaGenerator.java
```diff
@@ -1,4 +1,933 @@
...
+        while (sorted.hasNext()) {
+            Entry<String, TypeDef> prop = sorted.next();
+            TypeDef typeObj = prop.getValue();
+
+            JavaGenerator.setImportOf(typeObj, imports);
+
+            String desc = null;
+            if (typeObj.doc != null) {
+                desc = typeObj.doc;
+                desc = Tools.formatComment(desc, TAB + "", true);
```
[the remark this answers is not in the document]
algonautshant: I guess residue from formatting using find/replace...

generator/src/main/java/com/algorand/sdkutils/generators/OpenApiParser.java
```diff
@@ -239,49 +120,58 @@ TypeDef getType(
...
+                    return new TypeDef("byte[]", "binary", "getterSetter", 
+                            propName, goName, desc, isRequired(prop));
                 }
             case "AccountID":
                 break;
             case "BlockCertificate":
             case "BlockHeader":
-                addImport(imports, "java.util.HashMap");
-                return new TypeDef("HashMap<String,Object>", type, propName, goName, desc, isRequired(prop));
+                return new TypeDef("HashMap<String,Object>", type, "", propName, goName, desc, isRequired(prop));
```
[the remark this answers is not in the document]
algonautshant: Yes, it is left so. It is okay for now. It will be replaced with something more or less with something similar, which does not have the term "java" in it. Something like InterpretedType.

generator/src/main/java/com/algorand/sdkutils/generators/OpenApiParser.java
```diff
@@ -239,49 +120,58 @@ TypeDef getType(
...
             }
         }
         switch (type) {
         case "integer":
             String longName = asObject ? "Long" : "long"; 
-            return new TypeDef(longName, type, propName, goName, desc, isRequired(prop));
+            return new TypeDef(longName, type, "", propName, goName, desc, isRequired(prop));
         case "object":
         case "string":
```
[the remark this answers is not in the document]
algonautshant: Only used with ErrorResponse and HealthCheck. Fixed.

generator/src/main/java/com/algorand/sdkutils/utils/Tools.java
```diff
@@ -109,4 +116,53 @@ public static String getCamelCase(String name, boolean firstCap) {
         }
         return new String(newName, 0, n);
     }
+
+    // Imports are collected and organized before printed as import statements.
+    // addImports adds a needed import class. 
+    public static void addImport(Map<String, Set<String>> imports, String imp) {
```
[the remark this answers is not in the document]
algonautshant: Not sure it will "break". 
It depends how we want them organized. They will still be grouped together nicely, but the order might not be what you want. 
Anyway, complexity is contained within the Tools class.

generator/src/main/java/com/algorand/sdkutils/utils/Tools.java
```diff
@@ -109,4 +116,53 @@ public static String getCamelCase(String name, boolean firstCap) {
...
+    }   
+    
+
+    public static BufferedWriter getFileWriter(String className, String directory)  {
+        File f = new File(directory + "/" + className + ".java");
+        f.getParentFile().mkdirs();
+        BufferedWriter bw = null;
+        try {
+            bw = new BufferedWriter(new FileWriter(f));
+        } catch (IOException e) {
```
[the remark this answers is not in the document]
algonautshant: When it comes to exceptions, readability of the code is favored over exception handling. 
That is, exceptions are contained where they are generated, so the upstream code does not have to deal with them. 
They are also printed and execution continues. This helps debugging, since most exceptions do not impact the following processing. 
Clearly, this is not a production grade tool. Exceptions should not be expected.

generator/src/main/java/com/algorand/sdkutils/listeners/JavaGenerator.java
```diff
@@ -1,4 +1,933 @@
...
+            }
+            Tools.addImport(imports, "com.algorand.algosdk.crypto.Address");
+            return;
+        }
+    }
+
+    // Generate the enum file with accumulated and loaded enum definitions.
+    public void generateEnumClasses (String rootPath, String pkg)  {
+
+        BufferedWriter bw = Tools.getFileWriter("Enums", rootPath);
```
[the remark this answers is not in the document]
algonautshant: See the comment on exceptions overall.

generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
algonautshant: fixed.

generator/src/main/java/com/algorand/sdkutils/listeners/JavaGenerator.java
algonautshant: fixed.

generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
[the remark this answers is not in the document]
algorandskiy: StaleDelta -> StateDelta

generator/src/main/java/com/algorand/sdkutils/listeners/JavaGenerator.java
algonautshant: Simple: it means it has a getter/setter method. 
I agree it should not be here. It can be in the java code. Something for the future.

generator/src/main/java/com/algorand/sdkutils/listeners/JavaGenerator.java
algonautshant: I don't think all information should be passed to the listeners. 
The minimum amount of information necessary should be passed. And I think at this point more information is getting passed than it needs to be, but fixing this can be a simple, localized change for the future.

src/main/java/com/algorand/algosdk/v2/client/algod/GetBlock.java
algonautshant: :-) Good catch. 

:-( This is an existing bug. 

And now, all these files will lose a method, some after release... 

src/main/java/com/algorand/algosdk/v2/client/algod/AccountInformation.java
src/main/java/com/algorand/algosdk/v2/client/algod/GetApplicationByID.java
src/main/java/com/algorand/algosdk/v2/client/algod/GetAssetByID.java
src/main/java/com/algorand/algosdk/v2/client/algod/GetBlock.java
src/main/java/com/algorand/algosdk/v2/client/algod/GetPendingTransactionsByAddress.java
src/main/java/com/algorand/algosdk/v2/client/algod/PendingTransactionInformation.java
src/main/java/com/algorand/algosdk/v2/client/algod/WaitForBlock.java
src/main/java/com/algorand/algosdk/v2/client/indexer/LookupAccountByID.java
src/main/java/com/algorand/algosdk/v2/client/indexer/LookupAccountTransactions.java
src/main/java/com/algorand/algosdk/v2/client/indexer/LookupApplicationByID.java
src/main/java/com/algorand/algosdk/v2/client/indexer/LookupAssetBalances.java
src/main/java/com/algorand/algosdk/v2/client/indexer/LookupAssetByID.java
src/main/java/com/algorand/algosdk/v2/client/indexer/LookupAssetTransactions.java
src/main/java/com/algorand/algosdk/v2/client/indexer/LookupBlock.java

winder: I don't think thats true, otherwise the function wouldn't be green in this PR

generator/src/main/java/com/algorand/sdkutils/Generate.java
```diff
@@ -46,49 +47,44 @@ public static void Generate(
...
 
         if (goDirectory != null && !goDirectory.isEmpty()) {
-            g = new OpenApiParser(root, publisher);
             new GoGenerator(goDirectory, "indexer", publisher);
-        } else {
-            g = new OpenApiParser(root);
-        }
+        } 
 
+        new JavaGenerator(
```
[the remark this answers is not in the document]
algonautshant: fixed.

generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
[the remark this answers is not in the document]
algonautshant: fixed.

src/main/java/com/algorand/algosdk/v2/client/algod/GetBlock.java
```diff
@@ -32,6 +32,9 @@ public GetBlock(Client client, Long round) {
     }
 
     protected QueryData getRequestString() {
+        if (!qd.queries.containsKey("round")) {
```
winder, line 35: round isn't a query parameter, looks like this functionality needs to be extended to distinguish between path/body/query parameters.

src/main/java/com/algorand/algosdk/v2/client/common/AlgodClient.java
```diff
@@ -133,4 +137,39 @@ public GetPendingTransactions GetPendingTransactions() {
...
+     * /v2/teal/compile
+     */
+    public TealCompile TealCompile() {
+        return new TealCompile((Client) this);
+    }
+
+    /**
+     * /v2/teal/dryrun
+     */
+    public TealDryRun TealDryRun() {
```
winder, line 171: What do you think about having a constructor that accepts the body parameter? I guess we would need multiple for this (for the RawTransaction endpoint):

```
TealDryRun(){ return new TealDryRun((client) this); }
TealDryRun(DryRunParams params){ return new TealDryRun((client)this).params(params); }
```

winder: This probably isn't worth worrying about right now.

src/main/java/com/algorand/algosdk/v2/client/algod/GetBlock.java
algonautshant: you are right :-)
I fixed it here.

src/main/java/com/algorand/algosdk/v2/client/algod/GetBlock.java
algonautshant: Thanks for catching this. Yes, this is a regression.

PR #173, opened 2020-07-09
Implement TealSign and tests

algorandskiy:


src/main/java/com/algorand/algosdk/account/Account.java
```diff
@@ -531,6 +532,32 @@ public static SignedTransaction signLogicsigTransaction(LogicsigSignature lsig,
         }
     }
 
+    /**
+     * Creates Signature compatible with ed25519verify TEAL opcode from data and contract address (program hash).
+     * @param data byte[]
+     * @param contractAddress Address
+     * @return Signature
+     */
+    public Signature tealSign(byte[] data, Address contractAddress) throws NoSuchAlgorithmException, IOException {
```
winder, line 541: Could you use `ByteArrayOutputStream` instead of arrays and arraycopy?
There are a couple examples in the SDK, but it would be something like...

```
ByteArrayOutputStream baos = new ByteArrayOutputStream();
baos.write(PROGDATA_SIGN_PREFIX);
baos.write(rawAddress);
baos.write(data);
return this.rawSignBytes(baos.toByteArray());
```

algorandskiy: done

PR #175, opened 2020-07-10
Generator features and go enhancements

algonautshant:
* Parser support for Optional list of classes: only the listed classes are processed.
  This is useful when generated code will co-exist with hand written one.
  e.g. generating only application models for go.

* Fixed response type missing documentation.

* Optional name for go models file name (default: responsemodels.go)

* Enhanced main method for passing different args.

* Some go fixes (SignedTransaction, Account)

### DIFF
--- a/generator/src/main/java/com/algorand/sdkutils/generators/OpenApiParser.java
+++ b/generator/src/main/java/com/algorand/sdkutils/generators/OpenApiParser.java
@@ -26,9 +26,12 @@ public class OpenApiParser {
 
     protected JsonNode root;
     protected Publisher publisher;
-    protected final boolean javaMode;
 
     protected ObjectMapper mapper = new ObjectMapper();
+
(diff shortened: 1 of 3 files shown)


generator/src/main/java/com/algorand/sdkutils/generators/OpenApiParser.java
```diff
@@ -471,14 +487,20 @@ public void generateQueryMethods() throws IOException {
      */
     @Deprecated
     public OpenApiParser(JsonNode root) {
-        this.javaMode = true;
```
algorandskiy, line 474: How do we select Go or Java generator now?

algonautshant: we don't :-) why should we? it should always be the same :-)
if needed, can easily add another parameter to this, and not subscribe java or go listener.

PR #176, opened 2020-07-10
Fix  broken Generate.java

algonautshant:
When last commit added an additional parameter (models file name) to Go generator, this runner did not get updated.


algorandskiy:
Can we add a test that runs the generator using some spec file (mb the current algod.oas2.json is a good one for test purpose)?

algonautshant:
@winder added one. We are at a point where the runners are a fragmented for the different code generation tasks. 
Soon we should consolidate the pieces and get proper testing.

PR #177, opened 2020-07-11
Shant/applications ii

algonautshant:
This is the generator output replacing manual edits.
There are more comments added. Some may need a revision.


algonautshant:
> Looks good to me. Can you please elaborate which manual edits you refer to?

* Handling of in-body parameter
* setting msgPack format parameter

generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
```diff
@@ -47,6 +47,9 @@
     // constructing some functions.  
     private TreeMap<String, String> pathParameters;
 
+    // it is not expected to be more than 1 elt, but who knows...
+    private TreeMap<String, String> bodyParameter;
```
algorandskiy, line 51: haha

generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
```diff
@@ -303,13 +314,32 @@ private void addPathParameter(TypeDef type) {
...
+        // client functions
+        if (pathParameters.size() + bodyParameter.size() > 1) {
             clientFunction.append(", ");
         }
         clientFunction.append(propName + " " + gotype);
     }
 
     private void addQueryParameter(TypeDef type) {
+        // Do not expose format property
+        if (type.javaTypeName.equals("Enums.Format")) {
```
algorandskiy, line 339: not sure I understand how is this decision made

algonautshant: We are not exposing this parameter to the SDK. 
We silently set it to the default encoding.

generator/src/main/java/com/algorand/sdkutils/DoGenerate.java
```diff
@@ -0,0 +1,159 @@
...
+ * --filter
+ * 
+ *  Java:
+ *  -g java
+ *  --algodspec /Users/shantkarakashian/go/src/github.com/algorand/go-algorand/daemon/algod/api/algod.oas2.json
+ *  --indexerspec /Users/shantkarakashian/go/src/github.com/algorand/indexer/api/indexer.oas2.json
+ *  AccountStateDelta,Application,ApplicationLocalState,ApplicationLocalStates,ApplicationParams,ApplicationResponse,ApplicationStateSchema,ApplicationsResponse,CatchpointAbortResponse,CatchpointStartResponse,CompileResponse,DryrunRequest,DryrunResponse,DryrunSource,DryrunState,DryrunTxnResult,EvalDelta,EvalDeltaKeyValue,StateSchema,TealKeyValue,TealValue,TransactionApplication,GetApplicationByID,GetAssetByID,TealCompile,TealDryrun,LookupApplicationByID,SearchForApplications,LookupApplication
+ *  
+ */
+public class DoGenerate {
```
winder, line 36: We could be added as subcommands of `Main.java` as well, there's a build target so that you can run without the IDE:
```
~/generator$ mvn package
      ...
~/generator$  java -jar target/generator-1.0.0-jar-with-dependencies.jar -h
Usage: <main class> [options] [command] [command options]
  Options:
    -h, --help

    -s, --specfile
      Path to specfile used by generator.
  Commands:
    java      Generate the Java Client SDK.
      Usage: java [options]
        Options:
        * -n, --client-name
            Full path to the OpenAPI v2 spec file.
        * -cp, --common-package
            Package name to put at the top of generated client files.
        * -c, --common-path
            Path where generated client files will be placed.
          -h, --help

        * -mp, --model-package
            Package name to put at the top of generated model files.
        * -m, --model-path
            Path where generated model files will be placed.
        * -pp, --paths-package
            Package name to put at the top of generated paths files.
        * -p, --paths-path
            Path where generated paths files will be placed.
          -s, --specfile
            Path to specfile used by generator.
        * -t, --token-name
            Name of the token used for this application. i.e. 
            X-Algo-API-Token. 
          -tr, --token-required
            Whether or not a no-token version of the constructor should be 
            created. 
            Default: false

    responses      Generate response test file(s).
      Usage: responses [options]
        Options:
          -f, --filter
            Only generate response files for objects which contain the given 
            value as a substring.
          -h, --help

        * -o, --output-dir
            Location to place response objects.
        * -p, --prefix
            Prefix to resulting response files.
          -s, --specfile
            Path to specfile used by generator.
```

generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
winder: The SDKs have "real" transaction objects, so we opted to de-serialze the messagepack objects into those when possible.

src/main/java/com/algorand/algosdk/v2/client/algod/TealDryrun.java
```diff
@@ -23,9 +24,10 @@ public TealDryrun(Client client) {
 
     /**
      * Transaction (or group) and any accompanying state-simulation data.
+     * @throws JsonProcessingException
      */
-    public TealDryrun request(DryrunRequest request) {
-        addToBody(request);
+    public TealDryrun request(DryrunRequest request) throws JsonProcessingException {
+        addToBody(Encoder.encodeToMsgPack(request));
```
winder, line 30: :+1:

src/main/java/com/algorand/algosdk/v2/client/indexer/LookupAssetByID.java
```diff
@@ -16,6 +16,9 @@
 
     private Long assetId;
 
+    /**
+     * @param assetId
```
winder, line 20: Is there documentation we could add to the spec for these?

src/main/java/com/algorand/algosdk/v2/client/model/HealthCheck.java
```diff
@@ -11,7 +12,7 @@
 public class HealthCheck extends PathResponse {
```
winder, line 12: FYI, I noticed this when working on the response generator and didn't realize it was happening here as well. This isn't very important, but figured I should mention it since it would be nice to use `*Response` names everywhere eventually.

"Aliases" are not handled currently. We have `HealthCheckResponse` which references `HealthCheck`

I have some code in a branch that I should PR which adds a new `StructDef` field called `aliasOf`, and adds a new event like `StructDef(name="HealthCheckResopnse", aliasOf="HealthCheck")`

```
{
   "responses": {
      "HealthCheckResponse": {
         "description": "(empty)",
         "schema": {
            "$ref": "#/definitions/HealthCheck"
         }
      }
   }
   "definitions": {
     "HealthCheck": {
       "description": "A health check response.",
        "type": "object",
        "required": [
           "message"
        ],
        "properties": {
           "data": {
              "type": "object"
           },
           "message": {
              "type": "string"
           }
        }
     }
  }
}
```

winder: Also I'm not sure why I gave HealthCheck a data object, that seems like a mistake in the spec.

src/main/java/com/algorand/algosdk/v2/client/model/TransactionApplication.java
```diff
@@ -17,7 +17,6 @@
  */
 public class TransactionApplication extends PathResponse {
 
-    // TODO: Generator needs to be updated to setup the Address like this.
```
winder, line 20: I love this diff :)

src/main/java/com/algorand/algosdk/v2/client/model/HealthCheck.java
algonautshant: This is something we can think about in the future. Now, v2 is released and making changes won't go well I guess.

src/main/java/com/algorand/algosdk/v2/client/indexer/LookupAssetByID.java
algonautshant: Couldn't find it. 
What we can do is skip the comments for those.

generator/src/main/java/com/algorand/sdkutils/DoGenerate.java
algonautshant: Let's talk about this later. I need to understand some more to do it.

src/main/java/com/algorand/algosdk/v2/client/indexer/LookupAssetByID.java
winder: I think you can keep it, it's a good idea. The spec is probably just missing descriptions for the path parameters. These are sort of internal methods anyway.

PR #183, opened 2020-07-21
correct array of byte arrays used by application-args.

algonautshant:
1. Correct array of byte array type. It is used by application-args.

2. Fix for conflicting alias type classes 
No file should be generated when the type is an alias.
There are some alias types in one spec file, which
are in contradiction with real types in the other spec file.
We don't want the alias type file to corrupt the real type object.

### DIFF
--- a/generator/src/main/java/com/algorand/sdkutils/generators/OpenApiParser.java
+++ b/generator/src/main/java/com/algorand/sdkutils/generators/OpenApiParser.java
@@ -135,6 +135,7 @@ public class OpenApiParser {
             case "byte":
             case "base64":
             case "digest":
+            case "TEALProgram":
                 if (type.contentEquals("array")) {
                     type = prop.get("items").get("type").asText();
                     if (forModel == false) {
--- a/generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
+++ b/generator/src/main/java/com/algorand/sdkutils/listeners/GoGenerator.java
@@ -575,6 +575,7 @@ public class GoGenerator implements Subscriber {
             break;
 
         case "binary":
+        case "byte":
             goType = "[]byte";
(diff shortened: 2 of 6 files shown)


generator/src/main/java/com/algorand/sdkutils/generators/OpenApiParser.java
```diff
@@ -135,6 +135,7 @@ TypeDef getType(
             case "byte":
             case "base64":
             case "digest":
+            case "TEALProgram":
```
algorandskiy, line 138: it is weird, why do we need it? it is not in the spec

algonautshant: It is. In indexer spec file.

algorandskiy: I still do not understand why do we need it at all.

algonautshant: In the spec file, two format specifications are given: x-algorand-format and format:
        "approval-program": {
          "description": "\\[approv\\] approval program.",
          "type": "string",
          "format": "byte",
          "x-algorand-format": "TEALProgram"

At this point, the generator first looks at format, i.e., picks up "byte". Then, if x-algorand-format is specified, it overrides format, i.e.picks up "TEALProgram". Changing this behavior requires changes to other properties in the spec file.

Then, looks for a rule for TEALProgram. If one does not exist, then picks up type, which is "string".
This change adds a rule for TEALProgram, so that it does not go and pick "string". 

In my view, having TEALProgram is a good practice, since we are being more specific. 
What needs to be cleaned up is the use of specifications such as "byte", ore "base64", which are more ambiguous.

PR #195, opened 2020-09-09
Update client/generator to have a way to set custom headers.

winder:
Update the generator and regenerate clients. This also includes some recent fixes to the spec.


src/main/java/com/algorand/algosdk/v2/client/algod/GetVersion.java
```diff
@@ -0,0 +1,54 @@
...
+        resp.setValueType(Version.class);
+        return resp;
+    }
+
+    protected QueryData getRequestString() {
+        addPathSegment(String.valueOf("versions"));
+
+        return qd;
+    }
+}
```
algonautshant, line 54: Looks like the new line is missing here. 
Isn't this file generated?

algonautshant (approved):
Looks good.

src/main/java/com/algorand/algosdk/v2/client/algod/GetVersion.java
winder: Looks like that was missing, fixed now!

PR #197, opened 2020-11-13
New constructors for v2 and Indexer that also accepts tokenKey

fxgamundi:
Added new constructors to algodv2 and indexer classes that allows for a custom token key.
This allows connecting to APIs that use a different token keys from the default ones.
Given that the common Client class already allows this, the new constructors just calls super with the received parameters.


fxgamundi:
Not really sure why the build failed.
`Failed to connect to localhost/127.0.1.1:60002`

winder (approved):
LGTM

PR #201, opened 2021-01-07
Updated README example.

winder:
Use the Algod, kmd and Indexer clients to create a transaction on sandbox and check the result.


jasonpaulos (approved):
Looks good to me

PR #204, opened 2021-02-19
Clarify TransactionBuilder.suggestedParams

fabrice102:
Clarify that it overwrite parameters and should be called first.
This is in response to Discord message https://discord.com/channels/491256308461207573/631209325171441664/812081571053305906


winder:
@fabrice102 thanks for bringing this to my attention. I think meltie is right, it would be better if suggested params / loopup params only updated uninitialized values. What do you think? See #207

fabrice102:
@winder That's also a good option. The very small drawback is that it does break compatibility, but I really don't think it should be an issue.

PR #207, opened 2021-02-24
Don't override values with lookupParams/suggestedParams.

winder:
Alternative to #204 

If the developer directly sets a parameter, the suggestedParams helpers shouldn't override that value.


src/main/java/com/algorand/algosdk/builder/transaction/TransactionBuilder.java
```diff
@@ -90,7 +90,7 @@ final public Transaction build() {
 
     /**
      * Query the V1 REST API with {@link AlgodApi} for Transaction Parameters:
-     * Initialize fee, genesisID, genesisHash, firstValid, lastValid using {@link TransactionParams}.
+     * Initialize fee, genesisID, genesisHash, firstValid, lastValid if not already set using {@link TransactionParams}.
```
jasonpaulos, line 93:
```suggestion
     * Initialize fee, genesisID, genesisHash, firstValid, lastValid using {@link TransactionParams} if not already set.
```
This makes more sense to me.

jasonpaulos (commented):
Does a similar change need to be made for `public T suggestedParams(TransactionParametersResponse params)`?

PR #211, opened 2021-03-10
WIP / Proof of concept

winder:
I took a stab at re-implementing the java generator using templates.

I've moved most of the business logic into the templates by injecting helper functions. For example the OAP2 fields use kebab case throughout, and I created a kebab-to-camel function that is used in the templates. Other things like whether or not msgp is supported is going to be a little trickier (but I think not trickier than mixing callbacks and code-generation simultaneously)

### DIFF
--- a/generator/src/main/java/com/algorand/sdkutils/listeners/TemplateGenerator.java
+++ b/generator/src/main/java/com/algorand/sdkutils/listeners/TemplateGenerator.java
@@ -9,7 +9,7 @@ import com.algorand.sdkutils.utils.StructDef;
 import com.algorand.sdkutils.utils.TypeDef;
 import com.beust.jcommander.JCommander;
 import com.fasterxml.jackson.databind.JsonNode;
-import org.apache.logging.log4j.Level;
+import org.apache.commons.lang.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.velocity.Template;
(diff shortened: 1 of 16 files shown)


generator/src/main/java/com/algorand/sdkutils/listeners/TemplateGenerator.java
```diff
@@ -0,0 +1,256 @@
...
+            // TODO: Write to files instead of stdout.
+            System.out.println("====================");
+            System.out.println(StringUtils.capitalize(query.name) + ".java");
+            System.out.println("====================");
+            System.out.println(writer.toString());
+        }
+    }
+
+    @Override
+    public void terminate() {
```
winder, line 175: The template magic entrypoint. terminate is called after OpenAPI parsing completes at which point the template functions for models / queries / client is called. They all work slightly differently.

I'm still working on getting all of the parameters injected (i.e. the package).
Also a couple higher level knobs are needed for configuring filenames, and whether or not models are written 1 model per file, or all models in one file.

generator/src/main/java/com/algorand/sdkutils/listeners/TemplateGenerator.java
```diff
@@ -0,0 +1,256 @@
...
+        props.put("file.resource.loader.path", args.templatesDirectory.getAbsolutePath());
+        velocityEngine.init(props);
+
+        //Template modelTemplate = velocityEngine.getTemplate("model.vm");
+        //writeModelClass(modelTemplate);
+
+        Template queryTemplate = velocityEngine.getTemplate("query.vm");
+        writeQueryClass(queryTemplate);
+    }
+
```
winder, line 189: Below are the callbacks which trigger when the OpenAPIParser runs, they build up a series of objects that get passed into the templates.

generator/src/main/java/com/algorand/sdkutils/utils/QueryDef.java
```diff
@@ -29,4 +33,41 @@ public String toString() {
                 + ", description: " + this.description
                 + ", method: " + this.method;
     }
+
```
winder, line 36: Velocity needed getters to access these from the templates :(

winder:
Moved to separate repo

Issue #220, opened 2021-04-06
Error serializing application transaction using the Java SDK


winder:
@egieseke commented on [Mon Apr 05 2021](https://github.com/algorand/go-algorand-internal/issues/1112)

## Summary
JSON serialization issue for application transaction found in Java AlgoSDK version 1.5.1

## Scope/Requirements
Serialize and deserialize an application transaction in JSON.
 
## Urgency/Relative Priority
High.   Unable to send an application transaction through a REST API using the AlgoSDK Transaction class.

## Notes

In the below output, the OnCompletion code "apan" should be serialized as "UpdateApplicationOC" rather than the numeric enum value 4.

DEBUG [2021-04-05 03:55:49,017] io.dropwizard.jersey.jackson.JsonProcessingExceptionMapper: Unable to process JSON
! com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `com.algorand.algosdk.transaction.Transaction$OnCompletion` from String "4": 4 was not one of [NoOpOC, OptInOC, CloseOutOC, ClearStateOC, UpdateApplicationOC, DeleteApplicationOC]
!  at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 353] (through reference chain: com.businessapp.core.service.api.transaction.dto.SignedTransactionRequest["signedTransaction"]->com.algorand.algosdk.transaction.SignedTransaction["txn"]->com.algorand.algosdk.transaction.Transaction["apan"])
3:33 PM

[...]

PR #221, opened 2021-04-22
Add static qualifiers to json creators for onCompletion enum serialization.

egieseke:
Add static qualifiers to json creators for onCompletion enum serialization.

### DIFF
--- a/src/main/java/com/algorand/algosdk/transaction/Transaction.java
+++ b/src/main/java/com/algorand/algosdk/transaction/Transaction.java
@@ -1185,13 +1185,9 @@ public class Transaction implements Serializable {
             return null;
         }
 
-        public int getSerializedValue() {
-            return serializedValue;
-        }
-
         @JsonValue
         public int toValue() {
-            return getSerializedValue();
+            return serializedValue;
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -1185,9 +1185,13 @@ public OnCompletion String(int value) {
             return null;
         }
 
+        public int getSerializedValue() {
```
winder, line 1188: Why have a separate function here?

winder:
@egieseke thanks! Could you add the unit test from #220 to `TestTransaction.java` as well?

src/main/java/com/algorand/algosdk/transaction/Transaction.java
egieseke: Removed separate function.

egieseke: > @egieseke thanks! Could you add the unit test from #220 to `TestTransaction.java` as well?

Added unit test for Transaction and AppTransaction to TestTransaction.java.

egieseke: Removed this function and retested after removing JSONCreator annotation from string deserializer.

winder (approved):

PR #224, opened 2021-06-07
update langspec

shiqizng:
Update langspec.json to include latest TealV4 opcodes.


jannotti (approved):

src/main/resources/langspec.json
```diff
@@ -1 +1 @@
...
... =\u003e { 0 or 1}","Groups":["Byteslice Arithmetic"]},{"Opcode":170,"Name":"b%","Args":"BB","Returns":"B","Cost":20,"Size":1,"Doc":"A modulo B, where A and B are byte-arrays interpreted as big-endian unsigned integers. Panic if B is zero.","Groups":["Byteslice Arithmetic"]},{"Opcode":171,"Name":"b|","Args":"BB","Returns":"B","Cost":6,"Size":1,"Doc":"A bitwise-or B, where A and B are byte-arrays, zero-left extended to the greater of their lengths","Groups":["Byteslice Logic"]},{"Opcode":172,"Name":"b\u0026","Args":"BB","Returns":"B","Cost":6,"Size":1,"Doc":"A bitwise-and B, where A and B are byte-arrays, zero-left extended to the greater of their lengths","Groups":["Byteslice Logic"]},{"Opcode":173,"Name":"b^","Args":"BB","Returns":"B","Cost":6,"Size":1,"Doc":"A bitwise-xor B, where A and B are byte-arrays, zero-left extended to the greater of their lengths","Groups":["Byteslice Logic"]},{"Opcode":174,"Name":"b~","Args":"B","Returns":"B","Cost":4,"Size":1,"Doc":"A with all bits inverted","Groups":["Byteslice Logic"]},{"Opcode":175,"Name":"bzero","Args":"U","Returns":"B","Cost":1,"Size":1,"Doc":"push a byte-array of length A, containing all zero bytes","Groups":["Loading Values"]}]}
```
jdtzmn, line 1: nit: keep newline at end of file

jdtzmn (changes requested):

src/test/java/com/algorand/algosdk/logic/TestLogic.java
```diff
@@ -264,4 +264,55 @@ public void testCheckProgramTealV3() throws Exception {
             assertThat(valid).isTrue();
         }
     }
+
+    @Test
+    public void testCheckProgramTealV4() throws Exception {
+        assertThat(getEvalMaxVersion()).isGreaterThanOrEqualTo(4);
+        assertThat(getLogicSigVersion()).isGreaterThanOrEqualTo(3);
```
jasonpaulos, line 271: What's the difference between EvalMaxVersion and LogicSigVersion? Is there a reason we kept LogicSigVersion at 3, and is the expectation that we'll update this to 4 at some point?

@jannotti may be able to answer this.

jannotti:
```
	return &LanguageSpec{
		EvalMaxVersion:  logic.LogicVersion,
		LogicSigVersion: config.Consensus[protocol.ConsensusCurrentVersion].LogicSigVersion,
		Ops:             records,
	}
```
logic.LogicVersion is a "normal" Go variable, so it might be incremented was we work on, say, Teal 4, and even we release that, it will mean `goal` can compile v4,  but the network still won't evaluate it.  The consensus variable determines what the network will do.

So this test is running with a v4 compatible assembler, but a network that only takes v3 programs.  Is that the intent?

jannotti: Really I think it's just expressing that it wants the langspec.json file to have those values.  I don't know why.  I suspect it should not be testing the consensus version at all, as that is the consensus version when the langspec was written.  I don't understand why that's ever relevant to anyone. (Though I bet it is, I just can't think why right now!)

jasonpaulos:
```suggestion
        // TODO: check LogicSigVersion once v4 is supported by consensus
        // assertThat(getLogicSigVersion()).isGreaterThanOrEqualTo(4);
```
Thanks, that explanation makes sense.

It seems misleading to check that `LogicSigVersion` is greater to or equal to 3 in this test. I'd prefer one of:
1. Change the check to greater than or equal to 4, but comment it out for now (as in my suggested change).
2. Don't check `LogicSigVersion` at all, since it's dependent on consensus and the SDK doesn't need to care about that.

jannotti: You learn something new everyday - I didn't know you could do suggested changes! That's pretty nice. (though in this case I lean toward #2, unless we understand why it's worth testing the consensus value)

AyAggarwal (approved):

jasonpaulos (changes requested):
Mostly good, I'd just recommend to remove the check `assertThat(getLogicSigVersion()).isGreaterThanOrEqualTo(3);`

AyAggarwal:
I believe this is blocking my extraPage updates for this SDK. The only test i'm failing as of now is the new big program test, and its because of EvalMaxVersion

shiqizng:
@AyAggarwal since this change is blocking you, please merge this branch to yours and have this be in the same PR as your changes. The build for this will fail without the updated tests from your changes.

jdtzmn (approved):

jannotti (approved):

jasonpaulos (approved):
Looks good!

PR #226, opened 2021-06-11
url extender and unit test

AyAggarwal:


src/test/java/com/algorand/algosdk/transaction/TestTransaction.java
```diff
@@ -211,6 +212,21 @@ public void testAssetParamsValidation() throws Exception
             .hasMessageContaining("asset metadataHash '" +  badMetadataHash  + "' is not base64 encoded");
 
         assertThatThrownBy(() -> new AssetParams(
+                BigInteger.valueOf(100),
+                3,
+                false,
+                "tst",
+                "testcoin",
+                "websitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsitewebsite",
```
jdtzmn, line 220: This is significantly longer than 96 characters. Additionally, I'd recommend using something more concise and self-explanatory such as `'a'.repeat(97)`.

jdtzmn (changes requested):

src/test/java/com/algorand/algosdk/transaction/TestTransaction.java
AyAggarwal: good idea

jasonpaulos (commented):
This looks good, just one suggestion. Could you add a test that verifies a 96 character URL does not throw an error?

AyAggarwal:
> This looks good, just one suggestion. Could you add a test that verifies a 96 character URL does not throw an error?

yes, I'll make the passing case have a url of length 96

src/test/java/com/algorand/algosdk/transaction/TestTransaction.java
```diff
@@ -138,6 +138,7 @@ private void createAssetTest(int numDecimal, String goldenString) throws Excepti
         Address freeze = addr;
         Address clawback = addr;
         String metadataHash = "fACPO4nRgO55j1ndAK3W6Sgc4APkcyFh";
+        String url_96 = new String(new char[96]).replace("\0", "w");;
```
jdtzmn, line 141: Extra semicolon?

src/test/java/com/algorand/algosdk/transaction/TestTransaction.java
```diff
@@ -194,6 +195,8 @@ public void testAssetParamsValidation() throws Exception
         Address clawback = addr;
         String badMetadataHash = "fACPO4nRgO55j1ndAK3W6Sgc4APkcyF!";
         String tooLongMetadataHash = "fACPO4nRgO55j1ndAK3W6Sgc4APkcyFhfACPO4nRgO55j1ndAK3W6Sgc4APkcyFh";
+        String normalMetadataHash = "fACPO4nRgO55j1ndAK3W6Sgc4APkcyFhfACPO4";
+        String url_100 = new String(new char[100]).replace("\0", "w");;
```
jdtzmn, line 199: Extra semicolon?

jdtzmn (changes requested):

PR #227, opened 2021-06-17
Fee pooling SDK change

shiqizng:
This PR is to support creating transactions with a fee under mintxnfee since other txns in the group will be paying.


src/test/java/com/algorand/algosdk/cucumber/shared/TransactionSteps.java
```diff
@@ -24,6 +32,7 @@
     public Transaction applicationTransaction = null;
     public SignedTransaction signedTransaction = null;
 
+
```
jdtzmn, line 35: nit: extra line

src/test/java/com/algorand/algosdk/cucumber/shared/TransactionSteps.java
```diff
@@ -99,9 +108,9 @@ public void buildApplicationTransactions(String operation, Long applicationId, S
         if (StringUtils.isNotEmpty(appAccounts)) {
             builder.accounts(convertAccounts(appAccounts));
         }
-        if (fee != 0) {
-            builder.flatFee(fee);
-        }
+
+        builder.flatFee(fee);
```
jdtzmn, line 112: Not a huge deal, but I would move this up to line 93 so the property setters and conditionals are each grouped together.

src/test/java/com/algorand/algosdk/cucumber/shared/TransactionSteps.java
```diff
@@ -120,9 +129,30 @@ public void sign_the_transaction() throws NoSuchAlgorithmException {
...
+        assertThat(txn).containsKey("fee");
+    }
+
+    @Then("fee field not in txn")
+    public void  fee_field_not_in_txn() throws IOException {
+        if (signedTransaction == null) return;
+
+        byte[] encodedTxn = Encoder.encodeToMsgPack(signedTransaction);
+        Map<String,Object> sigtxn = Encoder.decodeFromMsgPack(encodedTxn,Map.class);
+        assertThat(sigtxn).isNotNull();
```
jdtzmn, line 148: Why is this check not in the `fee_field_is_in_txn()` method?

src/test/java/com/algorand/algosdk/unit/Rekeying.java
```diff
@@ -22,12 +27,17 @@
...
+    Address pk;
+    Transaction txn;
+    MultisigAddress msig;
 
     @Given("payment transaction parameters {int} {int} {int} {string} {string} {string} {int} {string} {string}")	
     public void payment_transaction_parameters(
             Integer fee, Integer fv, Integer lv, String gh,
             String to, String close, Integer amt, String gen, String note) {
 
+
```
jdtzmn, line 40: nit: extra line

src/test/java/com/algorand/algosdk/unit/Rekeying.java
```diff
@@ -70,4 +80,31 @@ public void the_signed_transaction_should_equal_the_golden(String golden) throws
     public void i_set_the_from_address_to(String fromAddress) {
         transactionBuilder.sender(fromAddress);
     }
+
+    @Given("multisig addresses {string}")
```
jdtzmn, line 84: Are these related to fee pooling? Or are they just sneaking into this PR for convenience?

jdtzmn (changes requested):

src/test/java/com/algorand/algosdk/cucumber/shared/TransactionSteps.java
shiqizng: Added.

src/test/java/com/algorand/algosdk/unit/Rekeying.java
shiqizng: these are steps for fee pooling tests, https://github.com/algorand/algorand-sdk-testing/blob/master/features/unit/feetest.feature#L29-L31. I had to put these in the Rekeying class because that's where the payment transaction builder is implemented. Cucumber doesn't allow me to duplicate the same definition in a different file.

src/test/java/com/algorand/algosdk/cucumber/shared/TransactionSteps.java
```diff
@@ -82,6 +90,8 @@ public void buildApplicationTransactions(String operation, Long applicationId, S
         }
 
         builder.sender(sender);
+        builder.flatFee(fee);
+
```
jdtzmn, line 94:
```suggestion
```

nit: extra line

src/test/java/com/algorand/algosdk/cucumber/shared/TransactionSteps.java
```diff
@@ -99,9 +109,7 @@ public void buildApplicationTransactions(String operation, Long applicationId, S
         if (StringUtils.isNotEmpty(appAccounts)) {
             builder.accounts(convertAccounts(appAccounts));
         }
-        if (fee != 0) {
-            builder.flatFee(fee);
-        }
+
```
jdtzmn, line 112:
```suggestion
```
nit: extra line

src/test/java/com/algorand/algosdk/unit/Rekeying.java
jdtzmn: Gotcha - that makes sense.

jdtzmn (approved):
2 extra lines but otherwise looks good!

src/test/java/com/algorand/algosdk/transaction/TestTransaction.java
```diff
@@ -81,6 +83,7 @@ public void testPaymentTransactionJsonSerialization() throws Exception {
                 .firstValid(301)
                 .lastValid(1300)
                 .genesisHash(new Digest())
+                .fee(0)
```
jasonpaulos, line 86: What happens if this line is not included? In order to not have a breaking change, we'll want the behavior for existing code like this to remain the same.

src/main/java/com/algorand/algosdk/builder/transaction/TransactionBuilder.java
```diff
@@ -77,13 +77,14 @@ final public Transaction build() {
             } catch (NoSuchAlgorithmException e) {
                 throw new UnsupportedOperationException(e);
             }
+            if (txn.fee == null || txn.fee == BigInteger.valueOf(0)) {
```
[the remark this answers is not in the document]
jasonpaulos: @winder could you weigh in?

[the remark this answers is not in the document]
shiqizng: I think == is right since txn.fee is an int and BigInteger.valueOf(0) is also an int.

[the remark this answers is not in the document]
jasonpaulos: I think `BigInteger.valueOf(0)` returns a BigInteger though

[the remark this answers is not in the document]
shiqizng: you're right.  txn.fee is also a BigInteger object. I'll update == to equals().

src/test/java/com/algorand/algosdk/transaction/TestTransaction.java
shiqizng: I updated the default fee in Transaction to 0 so it's consistent with the default in TransactionBuilder and removed this line.

src/test/java/com/algorand/algosdk/transaction/TestTransaction.java
```diff
@@ -766,4 +768,101 @@ public void EmptyByteArraysShouldBeRejected() throws Exception {
...
+                    .assetAmount(amountToSend)
+                    .fee(0)
+                    .firstValid(firstValidRound)
+                    .lastValid(lastValidRound)
+                    .genesisHash(gh)
+                    .assetIndex(assetIndex)
+                    .build();
+            assertThat(tx.fee).isEqualTo(1000);
+        }
+    }
```
jasonpaulos, line 866: Could you add more one test to verify that the default behavior when not calling `fee` or `flatFee` results in a transaction with a fee value of 1000? (That is the expected behavior right?)

src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -32,7 +32,7 @@
     @JsonProperty("snd")
     public Address sender = new Address();
     @JsonProperty("fee")
-    public BigInteger fee = Account.MIN_TX_FEE_UALGOS;
+    public BigInteger fee = BigInteger.valueOf(0);
```
jasonpaulos, line 35: nit: you could use `BigInteger.ZERO` instead of this constant.

jasonpaulos (changes requested):
Mostly good! I just want to verify that the default behavior when no fee is specified is the same. I'll approve after a test for that is added.

src/test/java/com/algorand/algosdk/transaction/TestTransaction.java
shiqizng: I actually changed default fee to 0. I'll switch it back to 1000.

src/main/java/com/algorand/algosdk/transaction/Transaction.java
winder: Change the default constructor to `public Transaction() {}`, it no longer needs to override the fee after this change.

src/test/java/com/algorand/algosdk/cucumber/shared/TransactionSteps.java
winder: One concern here is that in cucumber, there is no "not set" value so we've always used 0 instead. So in theory this could change the intent of some tests.

I suppose if all the tests pass we weren't using that feature for this field

src/test/java/com/algorand/algosdk/transaction/TestTransaction.java
```diff
@@ -766,4 +768,127 @@ public void EmptyByteArraysShouldBeRejected() throws Exception {
         assertThat(tx.note).isNull();
         assertThat(tx.lease).isNull();
     }
+
+    @Nested
```
winder, line 772: This is the first time I've never seen this feature, nice way to group similar tests! :+1:

jasonpaulos (approved):
Pending tests, looks good

PR #228, opened 2021-06-20
dynamicOpcode

AyAggarwal:


src/main/java/com/algorand/algosdk/logic/Logic.java
```diff
@@ -248,11 +248,11 @@ public static ProgramData readProgram(byte[] program, List<byte[]> args) throws
             }
             pc += size;
         }
-
-        if (cost > MAX_COST) {
-            throw new IllegalArgumentException("program too costly to run");
-        }
-
+        // costs calculated dynamically starting in v4
```
jasonpaulos, line 251: nit: could you fix the spacing here? Seems like a mix of tabs and spaces

jasonpaulos (approved):
Pending tests, looks good. Just one small comment about spaces

shiqizng (approved):
looks good to me.

PR #231, opened 2021-07-02
Added toSeed() method in Account.java

urajbhandari:


winder (approved):
Thanks for the contribution!

PR #236, opened 2021-07-07
set default header for base execute

shiqizng:


src/main/java/com/algorand/algosdk/v2/client/common/Client.java
```diff
@@ -103,6 +103,7 @@ public Response executeCall(QueryData qData, HttpMethod httpMethod, String[] hea
                     MediaType.parse("Binary data"),
                     qData.bodySegments.isEmpty() ? new byte[0] : qData.bodySegments.get(0));
             reqBuilder.post(rb);
+            reqBuilder.addHeader("Content-Type","application/x-binary");
```
winder, line 106:
```suggestion
                    MediaType.parse("application/x-binary"),
                    qData.bodySegments.isEmpty() ? new byte[0] : qData.bodySegments.get(0));
            reqBuilder.post(rb);
```

winder (approved):
This should get us by, thanks

PR #243, opened 2021-08-20
Added pretty print option to Object Mapper instance

cryptofervid:
It is difficult to interpret and read the stringified json output of Objects extending the PathResponse class as the output appears in a single line and is not formatted.

Added the option to use default pretty printer to the object mapper instance so that the outputs appear as formatted json strings.


winder:
@cryptofervid good idea, thanks!

PR #250, opened 2021-09-21
Feature/sign rekey lsig msig

### LINKED ISSUE #245: please add creating Account from privateKey

ahangsu:
This is the corresponding implementation and improvement from algorand/go-algorand-sdk#230.

This PR also added method for consttruction `Account` from `PrivateKey` and `KeyPair`, to resolve #245.

Details:
- MultiSig Changes:
  - `Account.signMultisigTransaction` can allow signing from any address.
  - `Account.mergeMultisigTransactions` now checks if multi-sig address and auth-address match.
- LogicSig Changes:
  - Introduced `LogicSigAccount`, which contains a `LogicSig` and (optional) a publicKey that signed it.
  - `LogicSig.verify` now only take address input when it's delegated to an account backed by one privateKey.
  - `Account.signLogicsigTransaction` can sign transaction from any address (take advantage of authAddr), except LogicSig is delegated to non-multisig account and sender is not the delegating account.
    (see algorand/go-algorand-sdk#230 for details)
  - `LogicSigAccount.signLogicSigTransaction` can sign transaction from any address without limitation.
- Minor:
  - `Account` has 2 more constructor, which allows constructing `Account` from `PrivateKey` and `KeyPair`.
  - Introduced a method to extract a `MultiSigAddress` from `MultiSigSignature`.

### DIFF
--- a/src/main/java/com/algorand/algosdk/account/Account.java
+++ b/src/main/java/com/algorand/algosdk/account/Account.java
@@ -14,6 +14,7 @@ import org.bouncycastle.asn1.ASN1Encodable;
 import org.bouncycastle.asn1.ASN1OctetString;
 import org.bouncycastle.asn1.pkcs.PrivateKeyInfo;
 
+import javax.annotation.Signed;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.math.BigInteger;
@@ -359,7 +360,7 @@ public class Account {
         SignedTransaction stx = new SignedTransaction(tx, mSig, txSig.transactionID);
         // if the transaction sender address is not multi-sig address
         // set the auth address as the multi-sig address
-        if (!tx.sender.toString().equals(from.toString()))
+        if (!tx.sender.equals(from.toAddress()))
             stx.authAddr = from.toAddress();
         return stx;
     }
@@ -380,6 +381,9 @@ public class Account {
                     tx.mSig.threshold != merged.mSig.threshold) {
                 throw new IllegalArgumentException("transaction msig parameters do not match");
             }
+            // check multi-sig address match
+            if (!tx.mSig.convertToMultisigAddress().equals(merged.mSig.convertToMultisigAddress()))
+                throw new IllegalArgumentException("transaction msig addresses do not match");
             // check that authAddr match
             if (!tx.authAddr.equals(merged.authAddr)) {
                 throw new IllegalArgumentException("transaction msig auth addr do not match");
@@ -551,6 +555,21 @@ public class Account {
 
     }
 
+    public static SignedTransaction signLogicTransactionWithAddress(LogicsigSignature lsig, Address lsigAddr, Transaction tx)
+            throws NoSuchAlgorithmException, IOException {
+        if (!lsig.verify(lsigAddr))
+            throw new IllegalArgumentException("verification failed on logic sig");
+
+        try {
+            SignedTransaction stx = new SignedTransaction(tx, lsig, tx.txID());
+            if (!stx.tx.sender.equals(lsigAddr))
+                stx.authAddr = lsigAddr;
+            return stx;
+        } catch (Exception ex) {
+            throw new IOException("could not encode transactions", ex);
+        }
+    }
+
     /**
      * Creates SignedTransaction from LogicsigSignature and Transaction.
      * LogicsigSignature must be valid and verifiable against transaction sender field.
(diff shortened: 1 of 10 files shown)


src/test/java/com/algorand/algosdk/account/TestAccount.java
```diff
@@ -334,4 +475,152 @@ void testToSeed() throws Exception {
         Account account = new Account(seed);
         assertThat(account.toSeed()).isEqualTo(seed);
     }
+
+    public static class TestLpgicSigTransaction {
```
jasonpaulos, line 479: spelling issue
```suggestion
    public static class TestLogicSigTransaction {
```

jasonpaulos (commented):
Nice work, the tests look great. I left some minor comments

src/main/java/com/algorand/algosdk/account/Account.java
```diff
@@ -514,23 +555,41 @@ public LogicsigSignature appendToLogicsig(LogicsigSignature lsig) throws Illegal
 
     }
 
+    public static SignedTransaction signLogicTransactionWithAddress(LogicsigSignature lsig, Address lsigAddr, Transaction tx)
+            throws NoSuchAlgorithmException, IOException {
```
winder, line 559: Do you need to declare `NoSuchAlgorithmException` here? It was probably added for creating `SignedTransaction`, but you convert it to an IOException.

src/main/java/com/algorand/algosdk/account/Account.java
```diff
@@ -514,23 +555,41 @@ public LogicsigSignature appendToLogicsig(LogicsigSignature lsig) throws Illegal
...
-        if (!lsig.verify(tx.sender)) {
-            throw new IllegalArgumentException("verification failed");
-        }
-
-        try {
-            return new SignedTransaction(tx, lsig, tx.txID());
-        } catch (Exception ex) {
-            throw new IOException("could not encode transactions", ex);
+    public static SignedTransaction signLogicsigTransaction(LogicsigSignature lsig, Transaction tx)
+            throws IllegalArgumentException, IOException, NoSuchAlgorithmException {
```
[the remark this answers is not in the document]
winder: Technically, yes. It would cause code to break if they upgrade. I'm not sure the new exception is needed though.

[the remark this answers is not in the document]
ahangsu: ah, so you mean I need to catch `NoSuchAlgorithmException` in the function, or someone else would be in trouble, right?

src/main/java/com/algorand/algosdk/account/Account.java
ahangsu: Yes, it comes from the `verify` in logicSigSignature, any modification needed on this?

jasonpaulos: It seems like we need to catch `NoSuchAlgorithmException` and wrap it in another exception to maintain backwards compatibility?

ahangsu: Yea, I think for both signLogicTransaction functions, we need to catch `NoSuchAlgorithmException`, but I am not sure about if anything need to be done on `verify` function.

src/main/java/com/algorand/algosdk/account/Account.java
```diff
@@ -514,22 +555,48 @@ public LogicsigSignature appendToLogicsig(LogicsigSignature lsig) throws Illegal
...
+            if (hasSig) {
+                lsigAddr = tx.sender;
+            } else if (hasMsig) {
+                lsigAddr = lsig.msig.convertToMultisigAddress().toAddress();
+            } else {
+                lsigAddr = lsig.toAddress();
+            }
+            return Account.signLogicTransactionWithAddress(lsig, lsigAddr, tx);
+        } catch (NoSuchAlgorithmException ex) {
+            throw new IllegalArgumentException("could not sign transaction", ex);
```
jasonpaulos, line 599: nit: can you change this to an `IOException` like before?

jasonpaulos (commented):
This looks good, I just have one suggestion.

src/main/java/com/algorand/algosdk/account/Account.java
ahangsu: Yea, need to check this

jasonpaulos (approved):

PR #255, opened 2021-09-29
Feature/abi encoding

### LINKED ISSUE #244: Add ABI encoding support
## Summary
Add support for encoding and decoding ABI values, as described in https://github.com/algorandfoundation/ARCs/pull/7

## Scope
Add ability to:
* Parse and serialize ABI type strings, e.g. `(uint64, byte[], address)`
* Encode and decode ABI values, using type information

ahangsu:
Corresponding implementation to https://github.com/algorand/go-algorand/pull/2807
- [X] serialization/deserialization of ABI types (and the corresponding testing)
- [ ] Definitions of ABI values (and the corresponding testing)
- [ ] Encoding/Decoding (with tests)

Closes #244


jasonpaulos (commented):
This is a good start, and I really like the approach of having a base `Type` and `Value` class and subclasses for every ABI type. It seems like it will be easier to add new ABI types in the future with this approach.

However, I would suggesting changing the subclass names from `<abi-type>T` and `<abi-type>V` to `Type<abi-type>` and `Value<abi-type>`, e.g. `TypeUint` and `ValueUint`.

src/main/java/com/algorand/algosdk/util/abi/values/ValueUfixed.java
```diff
@@ -0,0 +1,25 @@
...
+        this.value = value;
+    }
+
+    public ValueUfixed(int size, int precision, BigInteger value) {
+        this(new TypeUfixed(size, precision), value);
+    }
+
+    public byte[] encode() {
+        int ufixedSize = ((TypeUfixed) this.abiType).bitSize;
+        return Encoder.encodeUintToBytes(((BigInteger) this.value), ufixedSize);
```
algochoi, line 23: Should this be something like: 
```
int ufixedPrecision = ((TypeUfixed) this.abiType).precision;
return Encoder.encodeUintToBytes(((BigInteger) this.value * Math.pow(10, ufixedPrecision)), ufixedSize);
```

ahangsu: actually we are working over integer rather than rational, so we just encode the numerator over power of ten 😅

src/main/java/com/algorand/algosdk/util/abi/types/Type.java
```diff
@@ -0,0 +1,133 @@
...
+                else
+                    break;
+            } else {
+                until--;
+                break;
+            }
+        }
+        return until;
+    }
+}
```
algochoi, line 133: nit: Is there a linter/formatter for the Java SDK? Github is showing that some of the files need a newline at the end.

src/main/java/com/algorand/algosdk/util/abi/types/TypeArrayStatic.java
```diff
@@ -0,0 +1,49 @@
...
+        return Objects.hash(elemType, length);
+    }
+
+    public int byteLen() {
+        if (this.elemType instanceof TypeBool) {
+            List<Type> castedTupleElems = new ArrayList<>();
+            for (int i = 0; i < this.length; i++)
+                castedTupleElems.add(new TypeBool());
+            TypeTuple casted = new TypeTuple(castedTupleElems);
+            return casted.byteLen();
```
algochoi, line 45: Can we do a simpler calculation here instead of loops? e.g. `return ((int) Math.ceil(this.length/8.0))`

src/main/java/com/algorand/algosdk/util/abi/types/TypeUfixed.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Objects;
+
+public class TypeUfixed extends Type {
+    public final int bitSize;
+    public final int precision;
+
+    public TypeUfixed(int size, int precision) {
+        if (size < 8 || size > 512 || size % 8 != 0)
+            throw new IllegalArgumentException(
+                    "ufixed initialize failure: bitSize should be in [8, 512] and bitSize moe 8 = 0"
```
algochoi, line 12: nit: `mod` instead of `moe`, also consider using `equals` or `==` instead of `=`

src/main/java/com/algorand/algosdk/util/abi/values/Value.java
```diff
@@ -0,0 +1,171 @@
...
+        this.value = o.value;
+    }
+
+    public abstract byte[] encode();
+
+    public Value(byte[] encoded, Type abiType) {
+        this(Value.decode(encoded, abiType));
+    }
+
+    public static Value decode(byte[] encoded, Type abiType) {
```
algochoi, line 29: Can `decode` be declared as an `abstract` method here and be overriden in the child classes? It could help eliminate any switching or abiType checks, and we could probably combine the `TupleDecode` functions.

algochoi (commented):
This is looking very clean, just left some minor initial comments.

src/main/java/com/algorand/algosdk/util/abi/types/Type.java
ahangsu: have to manually add the lines, I doubt if there's code formatter/linter here in JAVA SDK...

src/main/java/com/algorand/algosdk/util/abi/values/Value.java
ahangsu: I had another idea. We can probably construct from the bytes and essential components for type.
In this case, the switching can be eliminated, and type checking is done while constructing type elements, the decoding is then proceed with new constructed type elements...

Seems to be fitting your thoughts?

algochoi: That's true -- I like the idea!

jasonpaulos (commented):
The more I think about it, the more the Python/Javascript approach of having no `Value` classes makes more sense to me. With the current implementation, this is the code needed to encode a value of type `(uint64,byte[])`:

```java
Value intValue = new ValueUint(76L);
Value[] byteValues = new Value[]{
    new ValueByte((byte)1),
    new ValueByte((byte)2),
    new ValueByte((byte)3),
};
Value byteArrayValue = new ValueArrayDynamic(byteValues, new TypeByte());

Value value = new ValueTuple(new Value[]{ intValue, byteArrayValue });

byte[] encoded = value.encode();
```

Comparatively, if there were no `Value` classes and the `Type` class had methods `public byte[] encode(Object value)` and `public Object decode(byte[])`, this is what it could look like:

```java
Type type = Type.fromString("(uint64,byte[])");

long intValue = 76;
byte[] byteValues = new byte[]{ 1, 2, 3 };

List<Object> tuple = List.of(intValue, byteValues);

type.encode(tuple);
```

The second seems significantly shorter, easier to write, and safer, since the type string would validate the Java values being encoded against the schema. Apologies for brining this up rather late, but what do you think?

(I also left some smaller comments)

ahangsu:
Removing `Value` class is fine, I agree with this decision, and it should be practical.
I had some further thinking: it seems golang supports `interface{}` type, which is similar in `Object` in Java. In this way, we can pass a list of interface type values into a encoding method, then infer the go-types and do the abi-type checking.

It seems to me that a rough structure of code change is shaped. Do we need to further reformat the code in go-algorand?

src/main/java/com/algorand/algosdk/util/abi/TypeArrayDynamic.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public byte[] encode(Object o) {
+        if (!(o instanceof Object[]))
+            throw new IllegalArgumentException("cannot infer type for abi dynamic array encode");
+        Object[] objArray = (Object[]) o;
+
+        byte[] castedEncode = Type.castToTupleType(objArray.length, this.elemType).encode(objArray);
+        byte[] lengthEncode = Encoder.encodeUintToBytes(BigInteger.valueOf(objArray.length), 2);
```
algochoi, line 38: Did we want to use a constant (ABI_LENGTH_BYTE_SIZE or something) instead of 2 for the length bytes?

src/main/java/com/algorand/algosdk/util/abi/TypeUint.java
```diff
@@ -0,0 +1,62 @@
...
+        return this.bitSize == ((TypeUint) o).bitSize;
+    }
+
+    @Override
+    public byte[] encode(Object obj) {
+        BigInteger value;
+        if (obj instanceof BigInteger)
+            value = (BigInteger) obj;
+        else if (obj instanceof Short)
+            value = new BigInteger(1, ByteBuffer.allocate(Short.SIZE / Byte.SIZE).putShort((short) obj).array());
```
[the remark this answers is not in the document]
ahangsu: I had another thought, is that if we eliminate all the negative values passed in, then go with your method of type casting to primitive numeric types, would that be fine?

src/main/java/com/algorand/algosdk/util/abi/Type.java
```diff
@@ -130,6 +131,21 @@ else if (currentIndex != 0 && delta < 0)
...
+            for (int i = 0; i < length; ++i)
+                outputArray[i] = Array.get(val, i);
+            return outputArray;
+        } else if (val instanceof List<?>)
+            return ((List<?>) val).toArray(new Object[0]);
+        else
+            throw new IllegalArgumentException("cannot infer type for unify array/list-like object to object array");
+    }
+
     public static byte[] getLengthEncoded(byte[] encoded) {
```
jasonpaulos, line 149: Can you make `getLengthEncoded`, `getContentEncoded`, and `castToTupleType` package local as well?

jasonpaulos (commented):
Looking good, just one final comment

src/main/java/com/algorand/algosdk/util/abi/Type.java
jasonpaulos: Once you add this, it looks good to merge

PR #262, opened 2021-10-12
Support parsing msgpack with numeric key

ahangsu:
- Upgrade Jackson msgpack library (with its dependencies)
- Updated `MultisigSubsig` Json serialization annotation


algochoi (approved):
Thank you to @jasonpaulos for helping resolving this! We should also add a cucumber test for `GetBlocks` to verify that responses are correct going forward.

ahangsu:
Much appreciation to @jasonpaulos !

PR #266, opened 2021-10-25
add new key reg txn field

shiqizng:
This PR adds a new field, stateProofKey, to key reg txn to match the changes in algod.


src/main/java/com/algorand/algosdk/crypto/Verifier.java
```diff
@@ -0,0 +1,47 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+import java.io.Serializable;
+
+/**
+ * A serializable representation of a state proof key.
+ */
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+public class Verifier implements Serializable {
```
jasonpaulos, line 14: The name `Verifier` works in go-algorand because it's scoped in a package, so every reference to it must be `merklekeystore.Verifier`. For the SDKs, I think we'll want a more descriptive name. Maybe `MerkleVerifier` or `StateProofVerifier`?

src/main/java/com/algorand/algosdk/builder/transaction/KeyRegistrationTransactionBuilder.java
```diff
@@ -92,8 +101,11 @@ public T participationPublicKey(byte[] pk) {
      * @return This builder.
      */
     public T participationPublicKeyBase64(String pk) {
-        this.votePK = new ParticipationPublicKey(Encoder.decodeFromBase64(pk));
+        if(!pk.isEmpty()){
```
jasonpaulos, line 104: These empty checks should not be needed. And they are actually a breaking change, since now `.participationPublicKeyBase64("abc...").participationPublicKeyBase64("")` produces a different result.

src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -319,7 +330,7 @@ public Transaction(Address sender, BigInteger fee, BigInteger firstValid, BigInt
     public static Transaction createKeyRegistrationTransaction(Address sender, BigInteger fee, BigInteger firstValid,
                                                                BigInteger lastValid, byte[] note, String genesisID,
                                                                Digest genesisHash, ParticipationPublicKey votePK,
-                                                               VRFPublicKey vrfPK, BigInteger voteFirst,
+                                                               VRFPublicKey vrfPK, MerkleVerifier sprfpk, BigInteger voteFirst,
```
jasonpaulos, line 333: This function change is considered breaking. Since Java allows function overloading, you could make a new one with these argument as well, but it looks like this function is deprecated anyway in favor of `KeyRegistrationTransactionBuilder`, so it would be fine if you didn't change it at all.

src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -625,6 +640,7 @@ private Transaction(@JsonProperty("type") Type type,
                         // keyreg fields
                         @JsonProperty("votekey") byte[] votePK,
                         @JsonProperty("selkey") byte[] vrfPK,
+                        @JsonProperty("sprfkey") byte[] stateProofKey,
```
jasonpaulos, line 643: nonpart also needs to be added here

src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -1276,6 +1297,7 @@ public boolean equals(Object o) {
                 voteFirst.equals(that.voteFirst) &&
                 voteLast.equals(that.voteLast) &&
                 voteKeyDilution.equals(that.voteKeyDilution) &&
+                nonpart==that.nonpart &&
```
jasonpaulos, line 1300: State proof key should also be compared

src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -729,10 +747,12 @@ public Transaction(
                         // keyreg fields
                         ParticipationPublicKey votePK,
                         VRFPublicKey vrfPK,
+                        MerkleVerifier stateProofKey,
                         BigInteger voteFirst,
                         BigInteger voteLast,
                         // voteKeyDilution
                         BigInteger voteKeyDilution,
+                        boolean nonpart,
```
jasonpaulos, line 755: Adding new args to this constructor breaks backwards compatibility. Unfortunately we didn't notice that when we added a new option for extra pages, but we shouldn't make that mistake again.

You could make a new constructor with the new args, or add them to this constructor at the end and give them a default value to maintain backwards compatibility. But probably the simplest solution is to not add them at all, since the main purpose of this constructor seems to be to make the deprecated txn constructor methods here work, and if you don't change `createKeyRegistrationTransaction`, there's no need to change this.

src/main/java/com/algorand/algosdk/v2/client/model/ErrorResponse.java
```diff
@@ -12,7 +11,7 @@
 public class ErrorResponse extends PathResponse {
 
     @JsonProperty("data")
-    public HashMap<String,Object> data;
+    public String data;
```
jasonpaulos, line 14: This seems like a breaking change, we should look into what spec change caused this

src/main/java/com/algorand/algosdk/builder/transaction/KeyRegistrationTransactionBuilder.java
```diff
@@ -221,4 +231,49 @@ public T voteKeyDilution(Long voteKeyDilution) {
...
+        return (T) this;
+    }
+
+    /**
+     * Set the stateProofKey value. This is used to identify a state proof
+     * @param stprf The stateProofKey value.
+     * @return This builder.
+     */
+    public T stateProofKeyBase64(String stprf) {
+        if(!stprf.isEmpty()){
```
shiqizng, line 261: I think I need to keep this check here so that stateProofKey can be an optional field by setting it to null. If not set to null, an empty stateProofKey is included in the txn as, `"sprfkey": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=="`

jasonpaulos (commented):
Some minor comments, plus I made #270 to fix the empty key getting encoded

Makefile
```diff
@@ -1,8 +1,8 @@
...
+	mvn test -Dcucumber.filter.tags="@unit.offline or @unit.algod or @unit.indexer or @unit.rekey or @unit.indexer.rekey or @unit.transactions or @unit.transactions.keyreg or @unit.responses or @unit.applications or @unit.dryrun or @unit.tealsign or @unit.responses.messagepack or @unit.responses.231 or @unit.responses.messagepack.231 or @unit.feetest or @unit.indexer.logs"
 
 integration:
-	mvn test -Dcucumber.filter.tags="@algod or @assets or @auction or @kmd or @send or @template or @indexer or @rekey or @applications.verified or @applications or @compile or @dryrun or @indexer.applications or @applications.evaldelta or @indexer.231"
+	mvn test -Dcucumber.filter.tags="@algod or @assets or @auction or @kmd or @send @send.keyregtxn or @template or @indexer or @rekey or @applications.verified or @applications or @compile or @dryrun or @indexer.applications or @applications.evaldelta or @indexer.231"
```
jasonpaulos, line 5:
```suggestion
	mvn test -Dcucumber.filter.tags="@algod or @assets or @auction or @kmd or @send or @send.keyregtxn or @template or @indexer or @rekey or @applications.verified or @applications or @compile or @dryrun or @indexer.applications or @applications.evaldelta or @indexer.231"
```
Just missing an "or"

jasonpaulos (commented):
Looks good. Though once stateproofs are merged into master, I would like to do a quick review again

winder (approved):

PR #268, opened 2021-11-01
ABI Interaction Support for JAVA SDK

### LINKED ISSUE #264: ABI Interaction - off-chain programs invoking on-chain contracts
As described in [this design document](https://gist.github.com/jasonpaulos/a810abe7e86d43840d14445718565a9a), we build support in the SDK for off-chain programs invoking on-chain contracts.

ahangsu:
This PR enables developers to invoke on-chain contract methods using off-chain applications, as defined by ARC-0004 and following the design doc: https://gist.github.com/jasonpaulos/a810abe7e86d43840d14445718565a9a

- [x] Defines an object for an ABI method description and expose the signature and selector
- [x] Fix method signature parsing
- [x] Defines an object for an ABI interface description
- [x] Defines an object for an ABI contract description
- [x] `TransactionSigner` and other background types implementation
- [x] `AtomicTransactionComposer` implementation

closes #264

## Test Plan

- [x] Cucumber test in ABI Json
- [x] Cucumber test in `AtomicTransactionComposer`


src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,415 @@
...
+     * An error will be thrown if signing any of the transactions fails.
+     *
+     * @return an array of signed transactions.
+     */
+    public List<SignedTransaction> gatherSignatures() throws IOException, NoSuchAlgorithmException {
+        if (this.status.compareTo(AtomicTxComposerStatus.SIGNED) >= 0)
+            return this.signedTxns;
+
+        List<TransactionWithSigner> txnAndSignerList = this.buildGroup();
+        for (TransactionWithSigner tws : txnAndSignerList) {
```
algochoi, line 190: Ideally, I think we want to merge the `TransactionSigners` and only call `sign` once per signer. 

Also, I think we would like to have a separate temporary `List` for the `stxn` and assign it to `this` at the end, so that if a `sign` fails, we can just throw away the temporary List instead of reverting back `this.signedTxns` to keep the "atomic" property.

algochoi (commented):
Neat implementation!

src/main/java/com/algorand/algosdk/abi/ABIType.java
```diff
@@ -78,16 +78,16 @@ public static Type fromString(String str) {
...
-                tupleTypes.add(Type.fromString(subStr));
+                tupleTypes.add(ABIType.Of(subStr));
             return new TypeTuple(tupleTypes);
         } else {
             throw new IllegalArgumentException("Cannot infer type from the string: " + str);
         }
     }
 
-    private static class Segment {
+    public static class Segment {
```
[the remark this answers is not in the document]
ahangsu: nice catch. I don't even remember why I made it public

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,415 @@
...
+        return this.transactionList.size();
+    }
+
+    /**
+     * Create a new composer with the same underlying transactions. The new composer's status will be
+     * BUILDING, so additional transactions may be added to it.
+     */
+    public AtomicTransactionComposer cloneAtomicTxnComposer() throws IOException {
+        AtomicTransactionComposer cloned = new AtomicTransactionComposer();
+        cloned.signedTxns = new ArrayList<>();
```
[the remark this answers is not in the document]
ahangsu: Good question, I am in some doubt about Java's deep copy mechanism, and I might overworked in this function. I think folks can discuss what should (or should not) be done on this thread '_>'

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,415 @@
...
+
+    /**
+     * Add a smart contract method call to this atomic group.
+     *
+     * An error will be thrown if the composer's status is not BUILDING, if adding this transaction
+     * causes the current group to exceed MAX_GROUP_SIZE, or if the provided arguments are invalid
+     * for the given method.
+     */
+    public void addMethodCall(MethodCallOption methodCall) {
+        if (this.status.compareTo(AtomicTxComposerStatus.BUILDING) != 0)
```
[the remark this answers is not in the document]
ahangsu: I am unaware of the `equals` method can be used. 
In the previous use cases we suppose to check the order of different enum field, so I stuck with `compareTo`. Will change soon.

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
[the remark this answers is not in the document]
algochoi: No worries, I was just curious because it seems like calling the constructor will automatically initialize the following:

`public List<SignedTransaction> signedTxns = new ArrayList<>();`

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,415 @@
...
+        for (MethodArgument arg : methodCall.methodArgs) {
+            if (arg instanceof ABIValue) {
+                abiValues.add((ABIValue) arg);
+            } else if (arg instanceof TransactionWithSigner) {
+                this.addTransaction((TransactionWithSigner) arg);
+            } else
+                throw new IllegalArgumentException("MethodArgument should only be ABI value or transaction with signer");
+        }
+
+        if (abiValues.size() > 14) {
```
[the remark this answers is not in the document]
ahangsu: but it seems the current implementation of `abiValues` only care about `ABIValue` instances, which can handle the issue of 10 `TxArgs` and 10 `ABITypes`. (it should just filter out the `TxnArg`s)

btw, should I check if the `abiValues` align with ABI types specified in `method` here? seems like I should?

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,415 @@
...
+    /**
+     * Add a transaction to this atomic group.
+     *
+     * An error will be thrown if the composer's status is not BUILDING, or if adding this transaction
+     * causes the current group to exceed MAX_GROUP_SIZE.
+     */
+    public void addTransaction(TransactionWithSigner txnAndSigner) {
+        for (byte b : txnAndSigner.txn.group.getBytes()) {
+            if (b != 0)
+                throw new IllegalArgumentException("Atomic Transaction Composer must not be non-zero");
```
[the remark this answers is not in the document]
ahangsu: Yes, probably clear the group field to be 0 in clone is correct, but according to design, if the group field is non-zero will incur error (though it should be overwritten later in assign group ID).

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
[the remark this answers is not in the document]
algochoi: Yes, I think your current implementation is correct. I think you can check, but it will error later when trying to encode the values anyway.

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,449 @@
...
+
+        public ReturnValue(String txID, byte[] rawValue, ABIValue value, Exception parseError) {
+            this.txID = txID;
+            this.rawValue = rawValue;
+            this.value = value;
+            this.parseError = parseError;
+        }
+    }
+
+    public static class TransactionSigner {
```
jasonpaulos, line 374: This would be a lot better as interface like I suggested here, but you may not have seen it: https://github.com/algorand/java-algorand-sdk/issues/264#issuecomment-942584208

The reason being we likely want to add more types of transaction signers in the future for things like KMD or WalletConnect/other wallet protocols. For this reason it's also probably better if TransactionSigner were declared in its own file instead of in this one.

Additionally if TransactionSigner were an interface, you could create anonymous implementations like this:

```java
    // inside of src/main/java/com/algorand/algosdk/account/Account.java
    public TransactionSigner getTransactionSigner() {
        Account self = this;
        return new TransactionSigner() {
            @Override
            public SignedTransaction[] signTransactions(Transaction[] txnGroup, int[] indexesToSign) throws Exception {
                SignedTransaction[] stxns = new SignedTransaction[indexesToSign.length];
                for (int i = 0; i < indexesToSign.length; i++) {
                    stxns[i] = self.signTransaction(txnGroup[indexesToSign[i]]);
                }
                return stxns;
            }
        };
    }
```

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,449 @@
...
+                return Account.mergeMultisigTransactions(multiStxns.toArray(new SignedTransaction[0]));
+            } else
+                throw new IllegalArgumentException("atomic transaction signer cannot infer sign transaction by who");
+        }
+    }
+
+    public interface MethodArgument {
+    }
+
+    public static class ABIValue implements MethodArgument {
```
jasonpaulos, line 429: I don't think there's any reason to have `MethodArgument` and `ABIValue` in this SDK. I left a similar comment on the Go SDK where I explained why: https://github.com/algorand/go-algorand-sdk/pull/258#discussion_r743308560

src/main/java/com/algorand/algosdk/transaction/MethodCallOption.java
```diff
@@ -0,0 +1,135 @@
...
+        this.lease = lease;
+        this.fv = fv;
+        this.lv = lv;
+        this.fee = fee;
+        this.flatFee = flatFee;
+        this.rekeyTo = rekeyTo;
+        this.signer = signer;
+    }
+
+    public static class MethodCallOptionBuilder {
```
jasonpaulos, line 45: Good choice on making a builder for this, it makes a lot of sense!

I'd suggest that `Builder` would be a better name, so to users it just looks like `MethodCallOption.Builder`

src/main/java/com/algorand/algosdk/transaction/MethodCallOption.java
```diff
@@ -0,0 +1,135 @@
...
+        public Long appID;
+        public Method method;
+        public List<AtomicTransactionComposer.MethodArgument> methodArgs;
+        public String sender, rekeyTo;
+        public Transaction.OnCompletion onCompletion;
+        public Byte[] note, lease;
```
jasonpaulos, line 51: nit: `byte[]` would make more sense for these types

src/main/java/com/algorand/algosdk/transaction/MethodCallOption.java
```diff
@@ -0,0 +1,135 @@
+package com.algorand.algosdk.transaction;
+
+import com.algorand.algosdk.abi.Method;
+import com.algorand.algosdk.algod.client.model.TransactionParams;
+
+import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MethodCallOption {
```
jasonpaulos, line 10: nit: `MethodCallParams` might be a better name?

jasonpaulos (commented):
Another suggestion: I think `ABIType.valueOf()` is more natural than `ABIType.Of()` for Java

src/main/java/com/algorand/algosdk/abi/Interface.java
```diff
@@ -0,0 +1,44 @@
...
+    @JsonCreator
+    public Interface(
+            @JsonProperty("name") String name,
+            @JsonProperty("methods") List<Method> methods
+    ) {
+        this.name = Objects.requireNonNull(name, "name must not be null");
+        if (methods != null) this.methods = methods;
+    }
+
+    @JsonIgnore
```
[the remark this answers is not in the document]
ahangsu: well, the weird thing in Jackson Json annotation is that, if you have `get`-alike function name, it will store unnecessary stuff in encoded Json string.

src/main/java/com/algorand/algosdk/transaction/MethodCallOption.java
ahangsu: Yes, I agree

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,449 @@
...
+        }
+        ByteBuffer bf = ByteBuffer.allocate(lengthEncoded);
+        for (byte[] subEncode : encodings)
+            bf.put(subEncode);
+        byte[] encoded = bf.array();
+        Response<PostTransactionsResponse> rPost = client.RawTransaction().rawtxn(encoded).execute();
+
+        if (!rPost.isSuccessful())
+            throw new Exception("transaction should be submitted successfully");
+
```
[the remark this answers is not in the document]
ahangsu: Uh... I guess the current version is using `ByteBuffer` to concatenate encoded **individual** group transactions. I learnt this from py-sdk.

One thing to be aware of: `Encoder.encodeToMsgPack` does not handle properly on a list of group txns, which will lead to some unexpected byte parsing error on the other end.

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
ahangsu: That is way better than my previous thoughts. I was constantly wondering how to get that integrate with other stuffs like KMD or wallet protocols, etc. This one seems more reasonable.

ahangsu: But another question arise: what if we want to use multi-sig, do we need to provide all the secret key and generate the `TransactionSigner`?

src/main/java/com/algorand/algosdk/abi/Interface.java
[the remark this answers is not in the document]
jasonpaulos: Ok, that makes sense, thanks!

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
jasonpaulos: Yes a multisig TransactionSigner will need a list of private keys to sign with. In the future we might want to revisit this to be more modular (maybe you can "join" basic TransactionSigners backed by a single private key into one multisig signer), but for simplicity I think this is the best approach for now

ahangsu: A follow-up question. I know we want to have *only one* call of sign from each `TransactionSigner`. Thus we might want to distinguish each interface value from another.

I think in this case we want to expose a method called `getAddress()`, which returns an address from each signer. In this way we can build a map of `TransactionSigner -> int[]`, which allows you to sign transaction group once from the signer.

jasonpaulos: That mostly sounds good, except I think it would be more general to use the `equals()` method to see if two `TransactionSigner`s are equivalent. The reason being there are some edge cases where two different signers need to sign for the same address (e.g. a delegated LogicSig can sign transactions for an account, but so can that account's secret key).

You could add the `equals` method as an abstract method on `TransactionSigner`, but then every instance would have to implement it. And currently since you are using anonymous instances, this would be hard to do.

So I'd probably suggest calling `equals()` from the `AtomicTransactionComposer` to check if `TransactionSigner`s are equivalent, but not actually implementing the `equals()` method on any instances. This results in Java comparing the objects by reference, which is fine for now, and something we can always improve later.

ahangsu: This sounds mostly reasonable to me, but I am concerned in the last part, since it is not necessarily true to have same instances of signer. I am wondering if I can use extension of interface like `comparable`, which might make things easier?

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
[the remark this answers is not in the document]
ahangsu: OK, now signedTxn should be automatically be an empty list.

src/main/java/com/algorand/algosdk/v2/client/model/TransactionSignatureMultisig.java
```diff
@@ -5,7 +5,7 @@
 import java.util.Objects;
 
 import com.algorand.algosdk.v2.client.common.PathResponse;
-import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.*;
```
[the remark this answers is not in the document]
ahangsu: yes it's from the generator, since we are going to import `JsonIgnore` too...

src/main/java/com/algorand/algosdk/transaction/MethodCalParams.java
```diff
@@ -0,0 +1,135 @@
+package com.algorand.algosdk.transaction;
+
+import com.algorand.algosdk.abi.Method;
+import com.algorand.algosdk.algod.client.model.TransactionParams;
+
+import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MethodCalParams {
```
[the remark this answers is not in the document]
ahangsu: uh oh...

src/test/java/com/algorand/algosdk/abi/TestInterface.java
```diff
@@ -0,0 +1,47 @@
...
+                        Arrays.asList(
+                                new Method.Arg("a", "uint64", "..."),
+                                new Method.Arg("b", "uint64", "...")
+                        ), null)
+        );
+    }
+
+    @Test
+    public void TestInterfaceClassMethods() throws IOException {
+        String testJSON = "{\"name\": \"Calculator\",\"methods\": [{ \"name\": \"add\", \"args\": [ { \"name\": \"a\", \"type\": \"uint64\", \"desc\": \"...\" },{ \"name\": \"b\", \"type\": \"uint64\", \"desc\": \"...\" } ] },{ \"name\": \"multiply\", \"args\": [ { \"name\": \"a\", \"type\": \"uint64\", \"desc\": \"...\" },{ \"name\": \"b\", \"type\": \"uint64\", \"desc\": \"...\" } ] }]}";
```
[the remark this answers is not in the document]
ahangsu: The code here takes null and generates `void` return type on default. I will change accordingly

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,443 @@
...
+        Transaction[] txnGroup = new Transaction[txnAndSignerList.size()];
+        for (int i = 0; i < txnAndSignerList.size(); i++)
+            txnGroup[i] = txnAndSignerList.get(i).txn;
+
+        for (TxnSigner txnSigner : transSignerToIndices.keySet()) {
+            List<Integer> indices = transSignerToIndices.get(txnSigner);
+            int[] indicesPrimitive = ArrayUtils.toPrimitive(indices.toArray(new Integer[0]));
+            SignedTransaction[] signed = txnSigner.signTxnGroup(txnGroup, indicesPrimitive);
+            for (int i = 0; i < indicesPrimitive.length; i++)
+                tempSignedTxns.set(indicesPrimitive[i], signed[i]);
```
algochoi, line 236: nit: we might want to check if `tempSignedTxns` has a `null` value in case some signer misbehaves

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,443 @@
...
+        List<TransactionWithSigner> txnAndSignerList = this.buildGroup();
+
+        HashMap<TxnSigner, List<Integer>> transSignerToIndices = new HashMap<>();
+        List<SignedTransaction> tempSignedTxns = new ArrayList<>();
+        for (int i = 0; i < txnAndSignerList.size(); i++) tempSignedTxns.add(null);
+
+        for (int i = 0; i < txnAndSignerList.size(); i++) {
+            TransactionWithSigner tws = txnAndSignerList.get(i);
+            if (transSignerToIndices.containsKey(tws.signer)) {
+                List<Integer> curr = transSignerToIndices.get(tws.signer);
```
algochoi, line 217: could this block be simplified to:

```suggestion
                transSignerToIndices.get(tws.signer).add(i);
```

algochoi (commented):
Looks pretty good to me! Very minor nits

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,439 @@
...
+            this.txIDs = txIDs;
+            this.methodResults = methodResults;
+        }
+    }
+
+    public static class ReturnValue {
+        public String txID;
+        public byte[] rawValue;
+        public Object value;
+        public String parseError;
```
jasonpaulos, line 411: Can you make this an `Exception` type again? That would be more useful, since it will contain the stacktrace of the error too

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,439 @@
...
+        for (int i = 0; i < this.signedTxns.size(); i++) {
+            if (this.signedTxns.get(i).tx.type.equals(Transaction.Type.ApplicationCall)) {
+                firstMethodCallIndex = i;
+                break;
+            }
+        }
+        firstMethodCallIndex = (firstMethodCallIndex == -1) ? 0 : firstMethodCallIndex;
+
+        PendingTransactionResponse txInfo =
+                Utils.waitForConfirmation(client, this.signedTxns.get(firstMethodCallIndex).transactionID, waitRounds);
```
jasonpaulos, line 315: It's only useful to wait on `firstMethodCallIndex` if you reuse the output of `Utils.waitForConfirmation` instead of calling `client.PendingTransactionInformation` for the first method call.

Also, right now `firstMethodCallIndex` is actually the index of the first app call transaction, not necessarily the first method call, since non-ABI app calls can be added with `addTransaction`.

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,439 @@
...
+        public ReturnValue(String txID, byte[] rawValue, Object value, String parseError) {
+            this.txID = txID;
+            this.rawValue = rawValue;
+            this.value = value;
+            this.parseError = parseError;
+        }
+    }
+
+    public interface TxnSigner {
+        int hashCode();
```
jasonpaulos, line 422: I think it would be more flexible if this type did not force subclasses to implement `hashCode` and `equals`. The default implementations of these methods should work fine for comparison by reference (which is fine for now), and subclasses can always choose to implement `hashCode` and `equals` if comparison by value is important to them.

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,439 @@
...
+
+        SignedTransaction signTxn(Transaction txn) throws NoSuchAlgorithmException, IOException;
+
+        SignedTransaction[] signTxnGroup(Transaction[] txnGroup, int[] indicesToSign)
+                throws NoSuchAlgorithmException, IOException;
```
jasonpaulos, line 427: Could you make this function throw `Exception` to be more general?

Also, is there a reason to keep the `signTxn` method? `signTxnGroup` is a more general form, and for simplicity I think we should have only one signing method for the time being.

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,439 @@
...
+    }
+
+    public static class TransactionWithSigner {
```
jasonpaulos, line 430: It would be great if you could move `TransactionWithSigner` and `TxnSigner` into their own files, since these types are not limited to being used only by `AtomicTransactionComposer`.

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -0,0 +1,439 @@
...
+                ));
+                continue;
+            }
+
+            byte[] abiEncoded = Arrays.copyOfRange(retLine, 4, retLine.length);
+            Object decoded = null;
+            String parseError = null;
+            try {
+                ABIType ABIType = this.methodMap.get(i).returns.parsedType;
+                decoded = ABIType.decode(abiEncoded);
```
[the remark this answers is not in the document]
ahangsu: my bad

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -63,15 +64,13 @@ public int getCount() {
      */
     public AtomicTransactionComposer cloneAtomicTxnComposer() throws IOException {
         AtomicTransactionComposer cloned = new AtomicTransactionComposer();
-        cloned.signedTxns = new ArrayList<>();
         for (Map.Entry<Integer, Method> entry : this.methodMap.entrySet())
             cloned.methodMap.put(entry.getKey(), new Method(entry.getValue()));
```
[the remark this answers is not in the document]
ahangsu: If `Method` objects are considered immutable, then it should definitely be fine to shallow copy.

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -110,32 +109,40 @@ public void addMethodCall(MethodCalParams methodCall) {
...
+        List<Object> methodArgs = new ArrayList<>();
+        List<ABIType> methodABIts = new ArrayList<>();
+
+        List<TransactionWithSigner> tempTransWithSigner = new ArrayList<>();
+
+        for (int i = 0; i < methodCall.method.args.size(); i++) {
+            Method.Arg argT = methodCall.method.args.get(i);
+            Object methodArg = methodCall.methodArgs.get(i);
+            if (argT.parsedType == null && methodArg instanceof TransactionWithSigner) {
+                tempTransWithSigner.add((TransactionWithSigner) methodArg);
```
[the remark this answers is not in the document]
ahangsu: Good catch

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
ahangsu: It sounds like it is an unnecessary move to use `firstMethodCallIndex`, assuming that any transaction in this transaction group is committed to the network at the same round.

Given the knowledge that we are holding all transaction id's, I suppose this is simplified to using first transaction id in signed transaction list?

jasonpaulos: Since the group is atomic, we can pass any of the TxIDs to `Utils.waitForConfirmation` to wait for the whole group to be committed. This function returns a `PendingTransactionResponse`, and since the execute function needs to obtain a `PendingTransactionResponse` for every method call in the group to examine logged values, the implementation could potentially reduce one API call if it passes a method call TxID to `Utils.waitForConfirmation`.

But this is just an optimization, so it's not necessary to implement. The current implementation is definitely acceptable.

jasonpaulos (approved):
Pending tests, looks good!

PR #274, opened 2021-11-10
Add WaitForConfirmation function

ahangsu:
Add wait for confirmation function for verifying transaction confirmation.


src/main/java/com/algorand/algosdk/v2/client/Utils.java
```diff
@@ -0,0 +1,45 @@
+package com.algorand.algosdk.v2.client;
+
+import com.algorand.algosdk.v2.client.common.AlgodClient;
+import com.algorand.algosdk.v2.client.common.Response;
+import com.algorand.algosdk.v2.client.model.NodeStatusResponse;
+import com.algorand.algosdk.v2.client.model.PendingTransactionResponse;
+
+public class Utils {
+    public static PendingTransactionResponse waitForConfirmation(AlgodClient client, String txID, Integer timeout)
```
jasonpaulos, line 9: Some nits:
* `int` instead of `Integer` would be better for `timeout`
* A name other than `timeout` would be better, since this value measures blocks, not time. That's why the other SDKs have opted for `waitRounds`.
* Also a docstring would be nice :grin:

jasonpaulos (approved):
Implementation looks good, just a few nits. Feel free to merge once they're addressed.

PR #275, opened 2021-11-23
Bug fix for `logs` on `PendingTransactionResponse`

ahangsu:
Bug fix for `PendingTransactionResponse`: could not populate `logs` array (of type `List<byte[]>`) from algod’s REST API.

Most of the changed files are generated from modified `generator`, the focus of this PR is `TestPendingTxnResponse`: we want to make sure we can parse `logs` out from `PendingTransactionResponse`.

### DIFF
--- a/src/main/java/com/algorand/algosdk/v2/client/model/Account.java
+++ b/src/main/java/com/algorand/algosdk/v2/client/model/Account.java
@@ -7,7 +7,7 @@ import java.util.Objects;
 
 import com.algorand.algosdk.crypto.Address;
 import com.algorand.algosdk.v2.client.common.PathResponse;
-import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.*;
 
 /**
  * Account information at a given round.
@@ -164,6 +164,7 @@ public class Account extends PathResponse {
      *   sig
      *   msig
      *   lsig
+     *   or null if unknown
      */
     @JsonProperty("sig-type")
     public Enums.SigType sigType;
--- a/src/main/java/com/algorand/algosdk/v2/client/model/Account.java
+++ b/src/main/java/com/algorand/algosdk/v2/client/model/Account.java
@@ -7,7 +7,7 @@ import java.util.Objects;
 
 import com.algorand.algosdk.crypto.Address;
 import com.algorand.algosdk.v2.client.common.PathResponse;
-import com.fasterxml.jackson.annotation.*;
+import com.fasterxml.jackson.annotation.JsonProperty;
 
 /**
  * Account information at a given round.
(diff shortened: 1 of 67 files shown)


jasonpaulos:
Could you add a unit test to verify that logs are getting decoded properly now?

Also, the code you modified is generated. It would be better if you modified the source template and then ran the generator again. The relevant lines: https://github.com/algorand/generator/blob/0e25557adaa9f2b353a3360e0fcd61ed2f03b4d7/src/main/java/com/algorand/sdkutils/listeners/JavaGenerator.java#L930-L957

As a side note, to avoid introducing a breaking change, would it be possible to keep the getter and setter methods, either with no annotation or with `@JsonIgnore`?

src/test/java/com/algorand/algosdk/util/TestPendingTxnResponse.java
```diff
@@ -0,0 +1,90 @@
...
+    )
+    public void decodeFromB64(String str) throws IOException {
+        byte[] decodedBytes = Encoder.decodeFromBase64(str);
+
+        PendingTransactionResponse decoded =
+                Encoder.decodeFromMsgPack(decodedBytes, PendingTransactionResponse.class);
+        assertThat(decoded.logs.size()).isEqualTo(1);
+
+        byte[] retLog = decoded.logs.get(0);
+        assertThat(retLog.length).isGreaterThan(4);
```
algorandskiy, line 76: please get rid of repeating the same numeric literal by introducing a named constant.

src/main/java/com/algorand/algosdk/v2/client/model/Account.java
```diff
@@ -7,7 +7,7 @@
 
 import com.algorand.algosdk.crypto.Address;
 import com.algorand.algosdk.v2.client.common.PathResponse;
-import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.*;
```
algorandskiy, line 10: it is considered a poor practice in Java to import '*'. Why did it happen at all?

algorandskiy: Linters actually may complain on both. Does it make the generator code more tricky?

algorandskiy: I checked JavaGenerator code.
Looks like `Tools.addImport(imports, "com.fasterxml.jackson.annotation. JsonIgnore")` can be moved into `getBase64EncodedArray` and to be conditional if needed since this code in `close()`:
```
        JavaGenerator.append(modelFileWriter, Tools.getImports(imports));
        JavaGenerator.append(modelFileWriter, currentModelBuffer);
```

ahangsu: great! I will go to change the code accordingly

ahangsu: It looks to me that it is working! I will soon have a small PR to resolve the wildcard-import in `generator`, would you like to take a pass?

algorandskiy (approved):

PR #277, opened 2021-12-01
Support Foreign objects as ABI arguments

### LINKED ISSUE #273: Support foreign objects as ABI arguments
## Summary
This is a follow up to #268, which added support for invoking ABI methods.

In addition to standard ABI type arguments, the SDK should support the special foreign object argument types `account`, `asset`, and `application`, [as described in ARC-4](https://github.com/algorandfoundation/ARCs/blob/6a9f91aed8068bbace872483a64cbf22d0e1c975/ARCs/arc-0004.md#foreign-apps-asas-and-accounts).

## Scope
Extend the `AtomicTransactionComposer.addMethodCall` function to support these new method argument types. At a high level it should:

1. Recognize when an argument is a special type or a standard ABI type. The special argument types are only valid as top-level arguments, so direct string comparison should be sufficient.
2. If a special type is present, insert the argument value into the appropriate foreign array, and use the array index of the value as the app call argument, treated as an ABI `uint8`.

ahangsu:
This PR adds support for Foreign accounts, apps, and assets as ABI arguments.

Currently, we are branching off from `feature/abi-interaction`, should be updated after that is merged to `develop`.

## TODO

- [x]  Accept foreign objects in method signature and convert values from method argument array
- [x] Check foreign array limits
- [x] "Compact foreign array": eliminate duplicated foreign objects (sender's addr/appID should be pointed at 0th apat/apas)
- [x] Allowing foreign arrays in `MethodCallParams`

Close #273, #282.

### DIFF
--- a/src/main/java/com/algorand/algosdk/abi/Contract.java
+++ b/src/main/java/com/algorand/algosdk/abi/Contract.java
@@ -7,25 +7,26 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class Contract {
     @JsonProperty("name")
     public String name;
-    @JsonProperty("appId")
-    public Integer appId;
+    @JsonProperty("networks")
+    public Map<String, AppID> networks;
     @JsonProperty("methods")
     public List<Method> methods = new ArrayList<>();
 
     @JsonCreator
     public Contract(
             @JsonProperty("name") String name,
-            @JsonProperty("appId") Integer appId,
+            @JsonProperty("networks") Map<String, AppID> networks,
             @JsonProperty("methods") List<Method> methods
     ) {
         this.name = Objects.requireNonNull(name, "name must not be null");
-        this.appId = Objects.requireNonNull(appId, "name must not be null");
+        this.networks = networks;
         if (methods != null) this.methods = methods;
     }
 
@@ -35,8 +36,8 @@ public class Contract {
     }
 
     @JsonIgnore
-    public Integer getAppId() {
-        return this.appId;
+    public Integer getAppIDbyHash(String genesisHash) {
+        return this.networks.get(genesisHash).appID;
     }
 
     @JsonIgnore
@@ -48,6 +49,29 @@ public class Contract {
     public boolean equals(Object o) {
         if (o == null || getClass() != o.getClass()) return false;
         Contract contract = (Contract) o;
-        return Objects.equals(name, contract.name) && Objects.equals(appId, contract.appId) && Objects.equals(methods, contract.methods);
+        if (!contract.networks.keySet().equals(this.networks.keySet())) return false;
+        for (String networkHash : this.networks.keySet()) {
+            if (!contract.getAppIDbyHash(networkHash).equals(this.getAppIDbyHash(networkHash)))
+                return false;
+        }
+        return Objects.equals(name, contract.name) && Objects.equals(methods, contract.methods);
+    }
+
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    static public class AppID {
+        @JsonProperty("appID")
+        public Integer appID;
+
+        @JsonCreator
+        public AppID(@JsonProperty("appID") Integer appID) {
+            this.appID = Objects.requireNonNull(appID, "appID must not be null");
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (o == null || o.getClass() != getClass()) return false;
+            AppID otherAppID = (AppID) o;
+            return Objects.equals(appID, otherAppID.appID);
+        }
     }
 }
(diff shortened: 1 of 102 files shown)


algochoi:
I think we probably want to add the following functionality as well (but not entirely sure yet):
1. Check foreign object limits (`account` array can have a maximum of 4, and the set of arrays can have a total maximum of 8)
2. "Compact" arguments -- if there are duplicate elements in the method args we want to have them all point to the same index in the foreign array. The sender and its own app id should also point to the 0th index of the `account` and `apps` array too I think.
3. Accept foreign objects in the `addMethodCall()`

ahangsu:
> * "Compact" arguments -- if there are duplicate elements in the method args we want to have them all point to the same index in the foreign array. The sender and its own app id should also point to the 0th index of the `account` and `apps` array too I think.

Just confirming: for eliminating duplicated elements in method args, this applies only to foreign objects, right?

src/main/java/com/algorand/algosdk/transaction/MethodCallParams.java
```diff
@@ -40,6 +78,26 @@ public MethodCallParams(Long appID, Method method, List<Object> methodArgs, Stri
...
+        this.foreignAssets = fAssets;
+        this.foreignApps = fApps;
+
+        Address senderAddr;
+        try {
+            senderAddr = new Address(this.sender);
+        } catch (Exception e) {
+            throw new IllegalArgumentException("sender mnemonic cannot generate an address");
+        }
+        this.foreignAccounts.remove(senderAddr);
```
algochoi, line 91: This is a nice optimization, but I'm wondering if it might mess up the existing indices:

e.g. `foreignAccounts = [senderAddr, addr2, addr3]` -> `[addr2, addr3]` then if the smart contract tries to access `foreignAccounts[3]`, it could panic and the user might not know what might be wrong?

src/test/java/com/algorand/algosdk/integration/AtomicTxnComposer.java
```diff
@@ -1,13 +1,16 @@
...
+import com.algorand.algosdk.crypto.TEALProgram;
 import com.algorand.algosdk.cucumber.shared.TransactionSteps;
 import com.algorand.algosdk.transaction.*;
 import com.algorand.algosdk.util.Encoder;
+import com.algorand.algosdk.util.ResourceUtils;
+import com.algorand.algosdk.util.SplitAndProcessABIArgs;
 import io.cucumber.java.en.Given;
 import io.cucumber.java.en.Then;
 import io.cucumber.java.en.When;
+import org.bouncycastle.util.Strings;
```
algochoi, line 13: Is the bouncycastle Strings being used?

src/test/java/com/algorand/algosdk/abi/TestMethod.java
```diff
@@ -6,42 +6,67 @@
 import org.junit.jupiter.api.Test;
 
 import java.io.IOException;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
```
algochoi, line 9: nit: I think the old import might still work (but if not, it might be better just to import the new packages individually anyway)

algochoi (commented):
Looks pretty good to me, just some small comments

src/test/java/com/algorand/algosdk/integration/AtomicTxnComposer.java
ahangsu: good catch, it is not used here

src/main/java/com/algorand/algosdk/transaction/MethodCallParams.java
ahangsu: This is an interesting example, ideally I think the foreign account array should not contain duplicated address, which means sender address should not be in there (even if it is not in the array but we still have an index for it).

My thoughts is that, the contract should not have access to index that lead to out of index error, since we populate the array index after we add the method call. I would like to leave this open for more discussion...

src/main/java/com/algorand/algosdk/abi/Contract.java
```diff
@@ -5,27 +5,29 @@
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
```
algochoi, line 8: nit: I think the old imports might still work here

src/main/java/com/algorand/algosdk/transaction/MethodCallParams.java
algochoi: My instinct is that this can be done at a lower level instead of at the SDK level, but I agree with your thoughts

jasonpaulos: I think this optimization is ok to make, since foreign objects are not guaranteed to be in a specific order in the ABI. And the main point of specifying additional foreign objects, as far as I can see, is to make sure they are "available" to the AVM during execution (for inner txns and other opcodes which require foreign references). So removing duplicates and the sender if present does not prevent that usage.

Though I would strongly recommend documenting that the foreign object arrays in `MethodCallParams` represent _additional_ foreign objects that the app should be able to access during execution, and that there are no guarantees on the ordering of the additional objects, or even if every additional object will be explicitly included in the app call transaction.

jasonpaulos (approved):
Looks good to me

algochoi (approved):

PR #281, opened 2021-12-08
Resolve #229 and #279 on upgrading jackson packages

ahangsu:
Resolve #229 and #279 on upgrading jackson packages.

Close #229 and #279.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,12 +60,12 @@
         <dependency>
           <groupId>com.fasterxml.jackson.core</groupId>
           <artifactId>jackson-core</artifactId>
-          <version>2.9.9</version>
+          <version>2.9.10</version>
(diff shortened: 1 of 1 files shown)


jasonpaulos:
According to https://search.maven.org/artifact/org.msgpack/jackson-dataformat-msgpack/0.9.0/jar, our version of `jackson-dataformat-msgpack` uses `jackson-databind` version  2.10.5.1, so maybe we should update to that if possible?

ahangsu:
AFAIK, if we want to up `jackson-databind` to 2.10.5.1, then `jackson-annotations` and `jackson-core` should be up to 2.10.0.

jasonpaulos (approved):
This looks good!

PR #283, opened 2021-12-09
Implementation for ABI spec change

### LINKED ISSUE #282: Implement ABI changes
Implement the changes described in https://github.com/algorandfoundation/ARCs/pull/53

Additionally, the following implementation details should change:
* The atomic transaction composer should not add a group ID if there's only 1 transaction in the group, since it's unnecessary
* The atomic transaction composer should check that the type of a transaction argument (e.g. `pay`, `axfer`, etc.) matches the ABI transaction argument type.

ahangsu:
## Tasks

- [x] Remove the `appId` field in favor of a new `networks` field in `Contract`.
  - [x] Waiting on cucumber test change, `Contract` construction is somewhat different.
- [x] If a method has exactly 15 arguments, the last argument is no longer required to be encoded in a tuple.
- [x] Clarify that in the case of multiple logged values with the special 4-byte return prefix, the last one is the return value.
- [x] The atomic transaction composer should not add a group ID if there's only 1 transaction in the group.
- [ ] The atomic transaction composer should check that the type of a transaction argument (e.g. pay, axfer, etc.) matches the ABI transaction argument type.

Closes #282.

### DIFF
--- a/src/main/java/com/algorand/algosdk/abi/Contract.java
+++ b/src/main/java/com/algorand/algosdk/abi/Contract.java
@@ -49,7 +49,12 @@ public class Contract {
     public boolean equals(Object o) {
         if (o == null || getClass() != o.getClass()) return false;
         Contract contract = (Contract) o;
-        return Objects.equals(name, contract.name) && Objects.equals(networks, contract.networks) && Objects.equals(methods, contract.methods);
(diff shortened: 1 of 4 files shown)


ahangsu:
Closing this PR since #277 contains all needed implementation.

PR #284, opened 2021-12-10
adding createDryrun method

barnjamin:
Untested, will work on that next but please take a look and make sure it's reasonable.

Seems like too many args for the method
Had to set the logic.StateSchema bytes/ints to public since that is what is used in Transaction class (tho we also have it in models?)


src/main/java/com/algorand/algosdk/v2/client/Utils.java
```diff
@@ -1,19 +1,162 @@
...
+import com.algorand.algosdk.v2.client.model.ApplicationStateSchema;
+import com.algorand.algosdk.v2.client.model.Account;
+import com.algorand.algosdk.v2.client.model.Asset;
 import com.algorand.algosdk.v2.client.model.NodeStatusResponse;
 import com.algorand.algosdk.v2.client.model.PendingTransactionResponse;
 
 public class Utils {
+
+    // https://github.com/algorand/go-algorand/blob/e466aa18d4d963868d6d15279b1c881977fa603f/libgoal/libgoal.go#L1089-L1090
+    public static Long defaultAppId = 1380011588L;
```
winder, line 28: This doesn't need to be public

src/main/java/com/algorand/algosdk/v2/client/Utils.java
```diff
@@ -1,19 +1,162 @@
...
+
+    /**
+     * Wait until a transaction has been confirmed or rejected by the network
+     * or wait until waitRound fully elapsed
```
winder, line 32: Wrong description.

It might be good to go into a lot of detail about how this function is used since it's a standalone feature-function that does a lot of stuff under the hood. If you want to go crazy, you could try embedding a code snippet ([with javadoc syntax](https://stackoverflow.com/a/542142)). I don't think we do that anywhere else though, so consider it optional.

src/main/java/com/algorand/algosdk/logic/StateSchema.java
```diff
@@ -12,10 +12,10 @@
 @JsonPropertyOrder(alphabetic = true)
 public class StateSchema implements Serializable {
     @JsonProperty("nui")
-    BigInteger numUint = BigInteger.ZERO;
+    public BigInteger numUint = BigInteger.ZERO;
 
     @JsonProperty("nbs")
-    BigInteger numByteSlice = BigInteger.ZERO;
+    public BigInteger numByteSlice = BigInteger.ZERO;
```
winder, line 18: You're using "ApplicationStateSchema", not "StateSchema", so this change probably isn't required.

winder (changes requested):
Looks like a helpful utility. I haven't used the DryRun feature, so I'll let @ahangsu respond to the correctness of how that is being constructed.

For my part, just some small cleanup requests.

src/main/java/com/algorand/algosdk/logic/StateSchema.java
barnjamin: For some reason the txn class uses StateSchema:  https://github.com/algorand/java-algorand-sdk/blob/develop/src/main/java/com/algorand/algosdk/transaction/Transaction.java#L142

When I wanted to copy it to the ApplicationStateSchema I wasn't able to access these fields without setting them to public. 

If I'm missing something please let me know

src/main/java/com/algorand/algosdk/v2/client/Utils.java
```diff
@@ -1,19 +1,161 @@
...
 import com.algorand.algosdk.v2.client.model.NodeStatusResponse;
 import com.algorand.algosdk.v2.client.model.PendingTransactionResponse;
 
 public class Utils {
+
+    // https://github.com/algorand/go-algorand/blob/e466aa18d4d963868d6d15279b1c881977fa603f/libgoal/libgoal.go#L1089-L1090
+    private static Long defaultAppId = 1380011588L;
+
+    /**
+     * Construct a DryruynRequest object from a set of transactions
```
winder, line 31: More detail please, this is included in the documentation (including in-line documentation provided by most Java IDEs) which is why I suggested that more is better:

https://algorand.github.io/java-algorand-sdk/com/algorand/algosdk/v2/client/Utils.html

I think it would be good to at least mention the function is going to query for asset/app creators as needed.

src/main/java/com/algorand/algosdk/v2/client/Utils.java
```diff
@@ -1,19 +1,161 @@
...
+     * 
+     * @param client           an Algod v2 client
+     * @param txns             the array of SignedTransactions that should be used
+     *                         to generate the DryrunRequest
+     * @param protocol_version The protocol version the dryrun should include
+     * @param latest_timestamp The latest timestamp the dryrun should include
+     * @param round            The agreement round or block height the dryrun should
```
winder, line 38: These parameters are for the other version of this function, could you make sure both functions are documented? I would write details in the version with 5 arguments and link to it from the version with 2 arguments, something like this:
> For additional details refer to {@link #createDryrun(AlgodClient, List<SignedTransaction>, String,
            Long, Long)}.

src/main/java/com/algorand/algosdk/v2/client/Utils.java
barnjamin: ah sure, was trying to keep it brief but makes sense to say what its actually doing

src/main/java/com/algorand/algosdk/v2/client/Utils.java
barnjamin: nice, thanks!

src/main/java/com/algorand/algosdk/v2/client/Utils.java
```diff
@@ -28,23 +28,41 @@
     private static Long defaultAppId = 1380011588L;
 
     /**
-     * Construct a DryruynRequest object from a set of transactions
+     * Construct a DryruynRequest object from a set of transactions. 
+     * A DryrunRequest is composed of static balance information.  This function uses the ApplicationCall transaction 
+     * parameters to infer what Application State and Account balance information to query using the client and adds it to the DryrunRequest object. 
+     * If foreign assets are passed, it will also add the creators balance information to the DryrunRequest. 
+     * For additional details refer to {@link #createDryrun(AlgodClient, List, String, Long, Long)}.
```
winder, line 35: No need for the link if you put the same description in both! Thanks for the extra details.

src/main/java/com/algorand/algosdk/v2/client/Utils.java
```diff
@@ -1,19 +1,179 @@
...
+     * 
+     * @param client           an Algod v2 client
+     * @param txns             the array of SignedTransactions that should be used
+     *                         to generate the DryrunRequest
+     * @return TransactionResponse of the confirmed transaction
```
winder, line 40: The return value needs to be updated also.

src/main/java/com/algorand/algosdk/v2/client/Utils.java
```diff
@@ -59,7 +58,7 @@ public static DryrunRequest createDryrun(AlgodClient client, List<SignedTransact
      * @param latest_timestamp The latest timestamp the dryrun should include
      * @param round            The agreement round or block height the dryrun should
      *                         include
-     * @return TransactionResponse of the confirmed transaction
+     * @return DryrunRequest of the confirmed transaction
```
winder, line 61: `@return DryrunRequest to be submitted to the TealDryrun endpoint`

barnjamin: jeeze ben

barnjamin:
While testing this with tealdbg && msgpack encoded DryrunRequest, I found that tealdbg is expecting the programs to be a byte array but I'm not sure I can force the output of `Encoder.encodeToMsgpack(drr)` to be the raw bytes.

Guessing here without digging too deep into the encoder/decoder stuff but I _think_ this decorator is used to flag the property we use when decoding and when its called, it automatically b64 encodes the underlying bytes:
https://github.com/algorand/java-algorand-sdk/blob/develop/src/main/java/com/algorand/algosdk/v2/client/model/ApplicationParams.java#L27


```
...
      "params": {
        "approval-program": "BSAIBAAKAWQCBgMmAwNyYWIBcANnb3YxGCMSQACyMRmBBRJAAKQxGSISQACXMRkhBRJAAIsxGSUSQACAMRkjEkAAAQA2GgCAB2RlcG9zaXQSQABiNhoAgAh3aXRoZHJhdxJAAEs2GgCABHN3YXASQAA4NhoAgAphZG1pbl9pbml0EkAAHzYaAIAMYWRtaW5fdXBkYXRlEkAAAQA2HAGIAfVCAC2IAgFCACeIAUBCACGIAOFCABuIADtCABUjQgARJUIADYgBuEIAB4gBskIAASVDMgokcAA1AzUEMgokcAA1BTUGNAM0BRBEKDQENAYKZyhkiTIEIQcSNjAAJBI2MAEhBBIQEDMAECEGEhA3ADAAMwEREhA3ADABMwIREhAzARAiEhAzARQyChIQMwERJBIQMwIQIhIQMwIUMgoSEDMCESEEEhAzAAAzAQASEDMBADMCABIQRCMoZTUBNQI0AUAABoj/bEIAAjQCNQAzAhI0AAszARISRLEishApZLIRMwESNAAKMwISCLISNhwAsgezJYkyBCEFEjYwACQSNjABIQQSEBAzABAhBhIQMwEQIhIQMwEUMgoSEDMBESlkEhBEsSKyECSyETMBEihkCrISMQCyFLOxIrIQIQSyETMBEihkCrISMQCyFLMliTYwADYwAQ1AAAaIAEZCAAOIAAGJMgQhBRI2MAAkEjYwASEEEhAQMwAQIQYSEDMBECISEDMBESQSEDMBEiMNEESxIrIQIQSyETMBEihkCrISsyWJMgQhBRI2MAAkEjYwASEEEhAQMwAQIQYSEDMBECISEDMBESEEEhAzARIjDRBEsSKyECSyETMBEihkC7ISsyWJIyplNQc1CDEANAdAAAUyCUIAAjQIEok1CYj/5EAABCNCAAUqNAlnJYmI/9RAAAQjQgADiAABibEhB7IQgA1EZW1vUG9vbFRva2VusiaAA2RwdLIlgYCU69wDsiIjsiMyCrIpMgqyKrMpMTxnJYk=",
...

**Comment by ahangsu** [comment]
@barnjamin I think your observation is correct, namely it automatically encodes the **raw bytes** in b64 format.
Minor question: do we want to enforce the program bytes to be exactly in raw byte format?

**Comment by barnjamin** [comment]
> do we want to enforce the program bytes to be exactly in raw byte format?

In order for it to be read by tealdbg when passing a msgp formated dryrun request, it must be the raw bytes.  The decoder logic of tealdbg tries to interpret the b64 string as the bytes of an assembled program. 

We need to either add something to this SDK to allow it or tweak the decoder in tealdbg to figure out it should try to b64 decode the bytes. 

Or  maybe there is another option?

**Review by ahangsu — APPROVED** [review]
LGTM

File: src/main/java/com/algorand/algosdk/v2/client/Utils.java [hunk 9]
```diff
@@ -1,19 +1,179 @@
 package com.algorand.algosdk.v2.client;
 
+import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Arrays;
```
ahangsu, line 6: nit: unnecessary import

src/main/java/com/algorand/algosdk/v2/client/Utils.java
```diff
@@ -1,19 +1,179 @@
...
+import com.algorand.algosdk.v2.client.model.ApplicationStateSchema;
+import com.algorand.algosdk.v2.client.model.Account;
+import com.algorand.algosdk.v2.client.model.Asset;
 import com.algorand.algosdk.v2.client.model.NodeStatusResponse;
 import com.algorand.algosdk.v2.client.model.PendingTransactionResponse;
 
 public class Utils {
+
+    // https://github.com/algorand/go-algorand/blob/e466aa18d4d963868d6d15279b1c881977fa603f/libgoal/libgoal.go#L1089-L1090
+    private static Long defaultAppId = 1380011588L;
```
ahangsu, line 28: nit: we can make this variable `final`:
```suggestion
    private static final Long defaultAppId = 1380011588L;
```

src/main/java/com/algorand/algosdk/v2/client/Utils.java
```diff
@@ -1,19 +1,179 @@
...
+
+            if (tx.foreignApps.size() > 0) {
+                apps.addAll(tx.foreignApps);
+            }
+
+            if (tx.foreignAssets.size() > 0) {
+                assets.addAll(tx.foreignAssets);
+            }
+
+            if (tx.accounts.size() > 0) {
```
ahangsu, line 127: I saw in implementation of createDryrun for other SDK that you add `sender` to `accts`, wondering do we also want to do this in Java, or `sender` is included in `tx.accounts` so we can omit this.

src/main/java/com/algorand/algosdk/v2/client/Utils.java
```diff
@@ -1,19 +1,179 @@
...
+            Response<Application> ar = client.GetApplicationByID(app).execute();
+            if(ar.isSuccessful()){
+                app_infos.add(ar.body());
+                accts.add(ar.body().params.creator.toString());
+            }
+        }
+
+        for (String acct : accts) {
+            Response<Account> ar = client.AccountInformation(new Address(acct)).execute();
+            if(ar.isSuccessful()){
```
ahangsu, line 152: one thing to make sure: for other SDK implementation, there's a decodeProgram involved here, this is not needed in Java case, right?

src/main/java/com/algorand/algosdk/v2/client/model/ApplicationParams.java
```diff
@@ -22,7 +22,6 @@
     public void approvalProgram(String base64Encoded) {
         this.approvalProgram = Encoder.decodeFromBase64(base64Encoded);
     }
-    @JsonProperty("approval-program")
```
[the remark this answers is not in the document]
barnjamin: I'll open one for generator, removing this annotation allows us to pass the bytes for the approval/clear program as bytes when we msgpack encode them rather than getting the b64 encoded version.

src/main/java/com/algorand/algosdk/v2/client/Utils.java
barnjamin: If I can edit the generator to allow raw bytes when we msgpack encode, this should work without decoding since the java decoder handles it for us.

src/main/java/com/algorand/algosdk/v2/client/model/ApplicationParams.java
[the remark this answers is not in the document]
ahangsu: Let me take a look soon...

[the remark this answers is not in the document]
ahangsu: Ping me once this is opened, and hopefully nothing's broken 😅

winder (approved):

PR #290, opened 2022-01-13
C2C Feature and Testing

### LINKED ISSUE #285: C2C feature testing and implementation
## Problem

We want to test C2C inner transaction call described in [c2c-feature-branch](https://github.com/algorand/go-algorand/tree/feature/contract-to-contract/data/transactions/logic#inner-transactions), and the discussion/design-doc is [here](https://docs.google.com/document/d/1uGM62VsN1F4aSTNg4t2q21JmuwS-OO-7bFY0VcaZcro/edit#).

We might want to have some deep-calls (a chain of inner-trans call) as testing method, which should be designed in SDK testing. The ideal result is that we can walk through inner transactions cascaded in tree-like structure.

ahangsu:
# Testing Contract to Contract Applications in the SDK

> DO NOT MERGE IT IN UNTIL `go-algorand/feature/c2c` IS MERGED INTO MASTER

closes #285 and #246 

## Summary of changes

- [x] `applications.verified` add test step for application address computable from appId hash.
- [x] ABI method return class add `PendingTransactionResponse` for C2C call result.
- [x] Compilation result byte string should match with expected.
- [x] Support method arg parsing for `ctxAppIdx`.
- [x] Support digging into ATC result with path (essentially we want to walk inside Transaction object).
- [x] Support digging into ATC call paths and check the group IDs.
- [x] Remove deprecated template feature for #246.
- [x] Revert `TEST_BRANCH` to `master` in `run_integration_tests.sh` (after `feature/c2c` is merged).


src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -417,36 +417,37 @@ public ExecuteResult execute(AlgodClient client, int waitRounds) throws Exceptio
...
+                        client.PendingTransactionInformation(signedTxns.get(i).transactionID).execute();
                 if (!resp.isSuccessful()) {
                     retList.add(new ReturnValue(
+                            signedTxns.get(i).transactionID,
                             null,
                             null,
-                            null,
-                            new Exception(resp.message())
+                            new Exception(resp.message()),
+                            resp.body()
```
jasonpaulos, line 430: Won't the call to `resp.body()` fail if the response was not successful?

src/test/java/com/algorand/algosdk/integration/Applications.java
```diff
@@ -137,6 +144,38 @@ public void waitForTransactionToBeConfirmed() throws Exception {
     public void rememberTheNewApplicatoinId() throws Exception {
         PendingTransactionResponse r = clients.v2Client.PendingTransactionInformation(txId).execute().body();
         this.appId = r.applicationIndex;
+        this.rememberedAppIds.add(this.appId);
+    }
+
+    @Given("I reset the array of application IDs to remember.")
+    public void i_reset_the_array_of_application_i_ds_to_remember() {
+        this.appId = 0L;
```
jasonpaulos, line 152: This should clear the `rememberedAppIds` list, not reset the most recent app ID

src/test/java/com/algorand/algosdk/integration/Stepdefs.java
```diff
@@ -743,7 +743,7 @@ public void readTxn(String encodedTxn, String num) throws IOException {
         String path = System.getProperty("user.dir");
         Path p = Paths.get(path);
         this.num = num;
-        path = p.getParent() + "/temp/raw" + this.num + ".tx";
+        path = p.getParent() + "/temp.json/raw" + this.num + ".tx";
```
jasonpaulos, line 746: Why make this change?

ahangsu: Good catch, I am unsure why this change happen on my commit...

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
ahangsu: No, it would return null. Shall I just say null here?

jasonpaulos: Ah ok. I think putting null would be clearer

jasonpaulos (approved):

src/main/java/com/algorand/algosdk/util/GenericObjToArray.java
```diff
@@ -0,0 +1,29 @@
+package com.algorand.algosdk.util;
```
tzaffi, line 1: 👍

src/test/java/com/algorand/algosdk/integration/AtomicTxnComposer.java
```diff
@@ -139,7 +151,172 @@ public void the_app_should_have_returned(String string) {
...
+                if (i == 0) {
+                    current = execRes.methodResults.get(path.get(i)).txInfo;
+                } else {
+                    current = current.innerTxns.get(path.get(i));
+                }
+            }
+            txInfoToCheck.add(current);
+        }
+
+        Digest groupID = new Digest();
```
tzaffi, line 312:
```suggestion
        Digest groupID;
```

tzaffi (commented):
Looks good. Just a few minor suggestions.

src/test/java/com/algorand/algosdk/util/ResourceUtils.java
```diff
@@ -10,4 +14,27 @@
...
     }
+
+    public static byte[] loadResource(String file) {
+        try {
+            return ResourceUtils.readResource(file);
+        } catch (Exception e) {
+            Assertions.fail("Unable to read file ('"+file+"') required by test: " + e.getMessage(), e);
+        }
+
+        throw new RuntimeException("Unknown error.");
```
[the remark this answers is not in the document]
ahangsu: This line is still reachable: `Assertions.fail` does not stop the execution, the related test will just fail.
I will not love it if it loads nothing and keeps running (e.g., return null bytes).

src/test/java/com/algorand/algosdk/integration/Applications.java
```diff
@@ -137,6 +144,38 @@ public void waitForTransactionToBeConfirmed() throws Exception {
...
+
+    @Given("I reset the array of application IDs to remember.")
+    public void i_reset_the_array_of_application_i_ds_to_remember() {
+        this.rememberedAppIds = new ArrayList<>();
+    }
+
+    @Given("I fund the current application's address with {int} microalgos.")
+    public void fundAppAccount(Integer amount) throws Exception {
+        Address appAddress = Address.forApplication(this.appId);
+        Address sender = base.getAddress(1);
```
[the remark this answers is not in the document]
tzaffi: 👍

PR #305, opened 2022-03-16
Dryrun printer

barnjamin:
Adding Dryrun trace printer to match other SDKs. Needs to be tested against cucumber tests


barnjamin:
I'd like https://github.com/algorand/java-algorand-sdk/pull/284 to be merged first

src/main/java/com/algorand/algosdk/v2/client/Utils.java
```diff
@@ -163,6 +178,151 @@ public static DryrunRequest createDryrun(AlgodClient client, List<SignedTransact
...
+    /**
+     * StackPrinterConfig contains configuration parameters for
+     * printing the trace from a DryrunTxnResult.
+     */
+    public static class StackPrinterConfig {
+        public int maxValueWidth = 0;
+        public boolean topOfStackFirst = false;
+    }
+
+    private static String truncate(String s, int maxWidth) {
```
winder, line 190: You can replace this with `StringUtils.abbreviate(s, maxWidth)`

https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#abbreviate-java.lang.String-int-

winder (changes requested):
Just one tiny request, otherwise LGTM.

src/main/java/com/algorand/algosdk/v2/client/Utils.java
jasonpaulos: There would still need to be logic to not call `StringUtils.abbreviate` if `maxWidth` is 0, but I agree with this

jasonpaulos (approved):
It would be nice to use the function Will suggested, but I have no strong feelings about it

src/main/java/com/algorand/algosdk/v2/client/Utils.java
barnjamin: I just tried but the `StringUtils.abbreviate` method doesn't work the same was as the other SDKs. 

 MaxWidth becomes maxwidth-3 to account for the ellipses it appends, whereas the other SDKs truncates to maxWidth (really maxValueWidth) then appends the ellipses. Even if I pass maxWidth+3 to the abbreviate method the results are different if the value length is between maxWidth-maxWidth+3 since it doesn't think it needs to do anything, so returns the original value with no truncation.

winder: Thanks, I see the case you're describing now. Probably not worth debating which way is more correct, so I'll go ahead and approve.

winder (approved):

PR #311, opened 2022-04-07
add error message with atomic transaction composer

Reguzzoni:
Add specific error message obtained by rPost transaction when It tried to submit an atomic transaction composer and It failed


jasonpaulos (approved):
Thanks for making this

PR #312, opened 2022-04-13
Added support for SHA256 TxnRoot header

Aharonee:
HashType enum was updated manually as it was not generated from the OpenAPI (due to the indexer oas2 spec file).


winder (changes requested):
Please update the spec so that the client code can be generated normally.

PR #318, opened 2022-05-11
Generate updated client API code

algoidurovic:
Depends on: https://github.com/algorand/go-algorand/pull/3957

### DIFF
--- a/src/main/java/com/algorand/algosdk/v2/client/algod/TealDisassemble.java
+++ b/src/main/java/com/algorand/algosdk/v2/client/algod/TealDisassemble.java
@@ -1,0 +1,68 @@
+package com.algorand.algosdk.v2.client.algod;
+
+import com.algorand.algosdk.v2.client.common.Client;
+import com.algorand.algosdk.v2.client.common.HttpMethod;
+import com.algorand.algosdk.v2.client.common.Query;
+import com.algorand.algosdk.v2.client.common.QueryData;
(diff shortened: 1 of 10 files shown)


ahangsu (commented):
~~The change seems right, but I didn't see a generator PR.~~

~~The error build message also seems off, the indexer build failed?~~

Now it seems we need to remove references to the old cost field in current implementation.

ahangsu (approved):
looks correct, waiting on Jason's review

src/main/java/com/algorand/algosdk/v2/client/algod/TealDisassemble.java
```diff
@@ -0,0 +1,68 @@
...
+public class TealDisassemble extends Query {
+
+    public TealDisassemble(Client client) {
+        super(client, new HttpMethod("post"));
+    }
+
+    /**
+     * TEAL program binary to be disassembled
+     */
+    public TealDisassemble source(byte[] source) {
```
jasonpaulos, line 26: Unlike the Go SDK this one seemed to pick up in the byte array input properly. But the base64 comment is still wrong on this class. For consistency I'd recommend getting rid of the TEAL disassembly changes to all SDKs, since it's a bit odd if only some of them support the feature.

jasonpaulos (commented):
I left a note about TEAL disassembly, but other than that this looks good

PR #321, opened 2022-05-20
Rerun code generation based on new models

Eric-Warehime:
Updated the generated files using the generator commit in 
https://github.com/algorand/generator/pull/31  
and the indexer spec in https://github.com/algorand/indexer/pull/1005

### DIFF
--- a/src/main/java/com/algorand/algosdk/v2/client/algod/GetProof.java
+++ b/src/main/java/com/algorand/algosdk/v2/client/algod/GetProof.java
@@ -5,6 +5,7 @@ import com.algorand.algosdk.v2.client.common.HttpMethod;
 import com.algorand.algosdk.v2.client.common.Query;
 import com.algorand.algosdk.v2.client.common.QueryData;
 import com.algorand.algosdk.v2.client.common.Response;
+import com.algorand.algosdk.v2.client.model.Enums;
 import com.algorand.algosdk.v2.client.model.ProofResponse;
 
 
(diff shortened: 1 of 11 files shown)


src/main/java/com/algorand/algosdk/v2/client/model/MiniAssetHolding.java
```diff
@@ -38,6 +38,20 @@ public String address() throws NoSuchAlgorithmException {
...
+     *   noop
+     *   optin
+     *   closeout
+     *   clear
+     *   update
+     *   update
+     *   delete
+     */
+    @JsonProperty("OnCompletion")
+    public Enums.OnCompletion onCompletion;
```
michaeldiamant, line 53: @Eric-Warehime and other reviewers - Sanity checking to confirm the `onCompletion` diff is expected?

From my inspection, it seems _unexpected_ though I still know little about API generation.
* Here's the indexer commit that I presume was referenced to generate the diff.    https://github.com/algorand/indexer/blob/0e409876e8def1de37791d6968320f0cf7da1ec5/api/indexer.oas2.json#L1637-L1671.
* It seems `MiniAssetHolding` does _not_ have `OnCompletion` as a field.  
* Curiously, it's the next type described.  Unclear if it's coincidental.

Eric-Warehime: Thanks for the callout, my previous generator changes were incorrectly associating enums with whichever model definition was alphabetically prior to it. 
Updates in the generator should now allow each language to take (or not take) a separate action on enum definitions allowing Java to store them in `Enums.java` regardless of their inclusion in paths, queries, etc.

michaeldiamant (approved):
@Eric-Warehime Nice - Unlike https://github.com/algorand/java-algorand-sdk/pull/320, the PR retains `Hashtype` while otherwise producing equivalent diffs.

PR #324, opened 2022-05-24
Introduce unified `MethodCallTransactionBuilder` & deprecate `MethodCallParams.Builder`

### LINKED ISSUE #323: MethodCallParams.Builder pulling v1 TransactionParams
### Subject of the issue

The method `setSuggestedParams()` of the Java object `MethodCallParams.Builder` used for ABI calls with ATC is pulling `TransactionParams` from v1 API instead of v2's `TransactionParametersResponse`. These objects are serialized differently, as a consequence it is not possible to call ABI methods with the Java SDK.

### Your environment

* Software version: `sandbox: 3.5.139117.dev`

### Steps to reproduce

    TransactionParametersResponse sp = algod.TransactionParams().execute().body();

    // create methodCallParams by builder (or create by constructor) for txntest method
    MethodCallParams.Builder methodCallParamsBuilderTxn = new MethodCallParams.Builder();
    // methodCallParams txn: commons params
    ...
    methodCallParamsBuilderTxn.setSuggestedParams(sp);

### Expected behaviour
Set suggested params pulled from algod v2 API as `TransactionParametersResponse` object

### Actual behaviour
Getting error:
`The method setSuggestedParams(TransactionParams) in the type MethodCallParams.Builder is not applicable for the arguments (TransactionParametersResponse)`

jasonpaulos:
This PR fixes a few related issues with `MethodCallParams.Builder`:
* It uses different setter names from the other transaction builders (e.g. `setAppID` vs `applicationId`)
* It lacks some convenience methods for setting fields (e.g. `setSuggestedParams` only accepted a v1 `TransactionParams` object, not a v2 `TransactionParametersResponse` object)
* It contains some minor logical errors, such as:
  * Always setting _both_ suggested parameters and first/last valid rounds and fees
  * Exposing members as public when not necessary

To remedy these, I've followed @winder's suggestion to make a new `MethodCallTransactionBuilder` class, similar to the other transaction builder classes. To make sure this class uses the same methods as the other builders, I've done the following:
* Make `TransactionBuilder` and `MethodCallTransactionBuilder` inherit from a new `TransactionParametersBuilder` class, which contains all common transaction parameters and their various setters.
* Introduce new interfaces `StateSchemaSetter`, `TEALProgramSetter`, and `ApplicationCallReferencesSetter`. These interfaces define setter methods for various application call transaction fields, and their purpose is to unify these setters between the application call transaction builders and the new `MethodCallTransactionBuilder`.

And to preserve backwards computability as much as is reasonable, the original `MethodCallParams.Builder` class still exists as an adapter between the preexisting interface and the new `MethodCallTransactionBuilder` class, though it is now deprecated.

Backwards incompatible changes are:
* `MethodCallParams.Builder`'s members are no longer public.
* `MethodCallParams`'s members have changed to reflect the internal types of the new `MethodCallTransactionBuilder`, and have become `final`. Its constructor has changed accordingly.

Closes #323


src/main/java/com/algorand/algosdk/builder/transaction/MethodCallTransactionBuilder.java
```diff
@@ -0,0 +1,171 @@
...
+import com.algorand.algosdk.transaction.MethodCallParams;
+import com.algorand.algosdk.transaction.Transaction;
+import com.algorand.algosdk.transaction.TxnSigner;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+
+@SuppressWarnings("unchecked")
+public class MethodCallTransactionBuilder<T extends MethodCallTransactionBuilder<T>> extends TransactionParametersBuilder<T> implements StateSchemaSetter<T>, TEALProgramSetter<T>, ApplicationCallReferencesSetter<T> {
```
michaeldiamant, line 16: @jasonpaulos I opened an optional child PR for your consideration:  https://github.com/algorand/java-algorand-sdk/pull/326.

Unless I misunderstand usage, I _think_ it's safe to remove the generic type for non-abstract members.  The PR illustrates the change.

Though as I'm writing, I see the other pre-existing, non-abstract classes still expose `T` (e.g. `AssetTransferTransactionBuilder`).  That's different than what I expected.  Since I already generated the PR, I'll leave as is for feedback.

jasonpaulos: The delegation vs inheritance change from that PR looks good -- I don't have a strong opinion here, but if delegation is preferred, that will solve the problem just as well.

As for the templating issue, the only technical drawback I'm aware of for removing the template is that `MethodCallTransactionBuilder` can't be subclassed. While that doesn't matter much currently, I'd rather not prohibit us from a subclass in the future, especially since all the other builders follow this same pattern.

michaeldiamant: @jasonpaulos Alright - I'll close #326.  Resolving.

I think subclassing a concrete type is _dangerous_.  However, that's the repo's convention, so let's keep as is.

michaeldiamant (approved):
@jasonpaulos Thanks for your effort to supply a fix here! ☕ 

It's the 1st time I've substantively interacted with java-algorand-sdk, so take my approval with a grain of salt.
* I think the PR definitively addresses the bug _and_ provides a solution for build conformance.
* Noting a passing concern - Generally, I am fearful of inheritance chains because they often increase fragility.  I'm _not_ requesting a change, but I feel obligated to vocalize the concern as a reviewer.  
  * I don't have enough experience with the repo to gauge the level of concern with `TransactionParametersBuilder`.
  * The inheritance is an expedient way to provide the same API.  From a brief scan, I don't immediately see what other abstract classes may appear.

michaeldiamant:
@jasonpaulos Approval note missed in previous comment:  Since I don't yet have a well-defined view here, I'm trusting the provided level of backwards compatibility is in conformance with repo conventions + norms.

src/main/java/com/algorand/algosdk/builder/transaction/MethodCallTransactionBuilder.java
winder: I don't remember the details, but I spent a lot of time back then messing around with different syntax to make sure the IDE was able to figure out exactly what options were valid.

i.e. when it wasn't working, if you were building an app call transaction and set a common field like "fee", autocomplete would no longer suggest "applicationID".

winder: Aside from that, I completely agree that this is pretty weird. If there is something simpler that would be nice.

jasonpaulos: We could get rid of the unchecked operations warnings with an approach like this: https://stackoverflow.com/a/5818701

But it doesn't change the templating syntax at all.

I'm generally in favor of a change like this, but we'd probably want to do it in a new PR, as the changes would be extensive.

winder: @jasonpaulos I agree, thats a great solution.

src/main/java/com/algorand/algosdk/builder/transaction/MethodCallTransactionBuilder.java
```diff
@@ -0,0 +1,171 @@
...
+            throw new IllegalArgumentException("extraPages must be an integer between 0 and 3 inclusive");
+        }
+        this.extraPages = extraPages;
+        return (T) this;
+    }
+
+    /**
+     * Build a MethodCallParams object.
+     */
+    public MethodCallParams build() {
```
winder, line 164: I don't think you should use `build` here, the other `build()` function is intentionally defined as `final`. When building a MethodCall there would be 2 build functions.

You should be able to move the `MethodCallParameters` logic into `applyTo`. This has the added benefit of reusing the fee setting logic directly rather than creating `MethodCallParams` -> applying the txn logic -> then calling into the TxnBuilder again to build the final app call transaction.

That said, I'm not well versed in the app transaction types and usage patterns. So maybe I'm misunderstanding the behavior.

jasonpaulos: Background: right now the flow of object creation is:
* `MethodCallTransactionBuilder` -> `MethodCallParams` -> `List<TransactionWithSigner>` (this list contains 0 or more "argument" transactions, then a final app call transaction)

The current code works because `MethodCallTransactionBuilder` does not inherit from `TransactionBuilder`, so it doesn't have a superclass with `build()` or `applyTo()` methods yet.

I do like this suggestion though, and it _almost_ works, but there are two things that prevent it:
1. The ultimate product of `MethodCallTransactionBuilder` is a `List<TransactionWithSigner>`, whereas `TransactionBuilder.build()` produces a `Transaction`. I suppose more methods could be added that return the additional information to go from a single `Transaction` to `List<TransactionWithSigner>`, but that doesn't seem optimal.
2. Unlike other transaction builders, there is a loss of information between `MethodCallTransactionBuilder` and its final product, `List<TransactionWithSigner>`; information about the method being called is lost, since only a hash of the method signature ends up in the final app call transaction. For this reason, `AtomicTransactionComposer` needs something that retains information about the method and can generate the final list of transactions, which is the role `MethodCallParams` fills right now.

We could get rid of `MethodCallParams`, but then we'd have to pass a `MethodCallTransactionBuilder` directly to `AtomicTransactionComposer.addMethodCall`. It seems a bit odd to me to call that object a builder when the user never calls a `.build()` on it, but I could be persuaded otherwise.

jasonpaulos: As a result of looking into this, I decided to move the method call transaction creation logic from the `AtomicTransactionComposer` to a new method, `MethodCallParams.createTransactions()`, which is more clear about the flow of creation I talked about above.

winder: Thanks for these details. Based on my limited understanding of the behavior, the changes look good to me.

winder (approved):

src/main/java/com/algorand/algosdk/builder/transaction/ApplicationCallReferencesSetter.java
```diff
@@ -0,0 +1,30 @@
+package com.algorand.algosdk.builder.transaction;
+
+import java.util.List;
+
+import com.algorand.algosdk.crypto.Address;
+
+public interface ApplicationCallReferencesSetter<T extends ApplicationCallReferencesSetter<T>> {
```
ahangsu, line 7: thanks for the discussion https://github.com/algorand/java-algorand-sdk/pull/324#discussion_r882194581, and now I understand the reason behind the syntax `T extends ...<T>` among all the txn builder.

ahangsu (approved):

PR #326, opened 2022-05-24
Delegate MethodCallParams.Builder to MethodCallTransactionBuilder

michaeldiamant:
Optionally proposes favoring composition over inheritance when adapting `MethodCallParams.Builder` to `MethodCallTransactionBuilder`.

The approach removes one level of generics definition.  Users of `MethodCallTransactionBuilder` no longer need to parameterize types.  From what I understood, there is no further specialization needed because `MethodCallTransactionBuilder` is a non-abstract class.

If I've misunderstood the intent or if we feel it's _not_ a net improvement, then feel welcomed to close the PR.

### DIFF
--- a/src/main/java/com/algorand/algosdk/transaction/MethodCallParams.java
+++ b/src/main/java/com/algorand/algosdk/transaction/MethodCallParams.java
@@ -9,7 +9,6 @@ import com.algorand.algosdk.crypto.TEALProgram;
 import com.algorand.algosdk.logic.StateSchema;
 import com.algorand.algosdk.v2.client.model.TransactionParametersResponse;
 
-import javax.swing.plaf.nimbus.State;
 import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


src/main/java/com/algorand/algosdk/transaction/MethodCallParams.java
```diff
@@ -9,6 +9,7 @@
 import com.algorand.algosdk.logic.StateSchema;
 import com.algorand.algosdk.v2.client.model.TransactionParametersResponse;
 
+import javax.swing.plaf.nimbus.State;
```
jasonpaulos, line 12: Accidental import?

jasonpaulos (commented):
The changes look good, thanks for making this PR. I just have a minor comment about a seemingly unnecessary import

src/main/java/com/algorand/algosdk/transaction/MethodCallParams.java
michaeldiamant: @jasonpaulos Yes, sorry about that.  Organized imports in https://github.com/algorand/java-algorand-sdk/pull/326/commits/8085a5d91ae540bb2182c3889ca286f32a0c8db5.

michaeldiamant:
Closing - Based on parent PR discussion, opting to keep as is to conform with repo conventions.

PR #334, opened 2022-06-03
adding method to abi result object

barnjamin:
Adds the method obj in the Return obj for abi contract calls


jasonpaulos (approved):
Looks good, just one nit

jasonpaulos (approved):

ahangsu (approved):

PR #337, opened 2022-06-09
adding tests and methods for getMethodByName

barnjamin:


src/test/java/com/algorand/algosdk/unit/ABIJson.java
```diff
@@ -149,4 +153,48 @@ public void the_deserialized_json_should_equal_the_original_contract_object() th
...
         assertThat(this.contract).isEqualTo(deserialized);
     }
+
+    @When("I append to my Method objects list in the case of a non-empty signature {string}")
+    public void i_append_to_my_method_objects_list_in_the_case_of_a_non_empty_signature(String methodsig) {
+        if(this.methods == null){
+            this.methods = new ArrayList<>();
+        }
+
+        if(!methodsig.equals("")){
```
ahangsu, line 163: nit: I guess you can use `isEmpty` method from java-string?

src/main/java/com/algorand/algosdk/abi/Interface.java
```diff
@@ -44,6 +45,31 @@ public Method getMethodByIndex(int index) {
...
+        if(methods.size()>1){
+            String[] sigs = new String[methods.size()];
+            for(int idx=0;idx<methods.size();idx++){
+               sigs[idx] = methods.get(idx).getSignature();
+            }
+            String found = StringUtil.join(sigs, ",");
+            throw new IllegalArgumentException(String.format("found %d methods with the same name: %s", methods.size(), found));
+        }
+
+        if(methods.size()==1){
```
ahangsu, line 66: should be `size() == 0`?

ahangsu (commented):
implementation looks good to me, we are waiting for a specific tag for cucumber test, right?

ahangsu (approved):

PR #341, opened 2022-07-01
Run non-Cucumber unit tests during make test

michaeldiamant:
Modifies `make test` to include non-Cucumber unit tests.

I'm unfamiliar with the history to gauge _why_ non-Cucumber unit tests are _not_ run during local test cycles.  I see the CI build _does_ run non-Cucumber unit tests.  In a future PR, I envision unifying local + CI builds.


michaeldiamant:
Closing - I misread the output.  Non-Cucumber unit tests _are_ run in `make test`.  Though local + CI builds still ought to be unified.

PR #336, opened 2022-06-03
Support for BoxReference Transaction array field.

algoidurovic:

PR #343, opened 2022-07-01
Boxes: Query algod Box by application ID and Box name

michaeldiamant:
Adds support to query algod for Box by application ID and Box name with unit + integration algorand-sdk-testing support.

Notes:
* Thanks to @algoidurovic for providing a base in #336.
* The PR makes no attempt to provide convenience methods for supplying an encoded box name. Convenience methods will be provided in a future PR.


algochoi (commented):
Looks good, just some questions for my understanding:
1. Is this code all generated? (Excluding the `makefile`)
2. Will there be any utility functions to convert into the `app-arg` format encoded name? Also the current cucumber tests only check for `b64:` prefixed encodings, so it may be prudent to check other tests if we choose to go down this path for Go/Java

algochoi (approved):
LGTM 📦

algoidurovic (approved):
Thanks for getting this up to date!

PR #346, opened 2022-07-07
Boxes:  Add convenience methods for encoding Box names

michaeldiamant:
Adds convenience methods to ensure proper Box name encoding when querying Boxes.

Details:
* Since `com.algorand.algosdk.v2.client.common.AlgodClient` is auto-generated, it's _not_ possible to provide manual extensions (e.g. expose another route).  So, the PR opts to provide utility methods that can be used by the library user.
* PR defines `BoxQueryEncoding` in `util` package following the repo's prevailing convenience method location.

### DIFF
--- a/src/main/java/com/algorand/algosdk/util/BoxQueryEncoding.java
+++ b/src/main/java/com/algorand/algosdk/util/BoxQueryEncoding.java
@@ -1,8 +1,8 @@
-package com.algorand.algosdk.v2.client.model;
+package com.algorand.algosdk.util;
 
 import com.algorand.algosdk.transaction.AppBoxReference;
 import com.algorand.algosdk.transaction.Transaction;
-import com.algorand.algosdk.util.Encoder;
+import com.algorand.algosdk.v2.client.model.Box;
 
 /**
  * BoxQueryEncoding provides convenience methods to String encode box names for use with Box search APIs (e.g. GetApplicationBoxByName).
--- a/src/main/java/com/algorand/algosdk/v2/client/model/BoxQueryEncoding.java
+++ b/src/main/java/com/algorand/algosdk/v2/client/model/BoxQueryEncoding.java
@@ -1,0 +1,29 @@
+package com.algorand.algosdk.v2.client.model;
+
+import com.algorand.algosdk.transaction.AppBoxReference;
+import com.algorand.algosdk.transaction.Transaction;
+import com.algorand.algosdk.util.Encoder;
+
+/**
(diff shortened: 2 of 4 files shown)


michaeldiamant:
> How do we feel about supplying this method as a member in AppBoxReference instead of its own class? 

@algochoi It's a reasonable question.  @algoidurovic raised a similar question in go-algorand-sdk.

* Here's my original thinking: https://github.com/algorand/go-algorand-sdk/pull/342#discussion_r916068414.
* Since you've both provided similar feedback, I'm on-board to make the change for `AppBoxReference` and `BoxReference`.
* Do you agree with retaining `Box`'s convenience method as is?  
  * I'd like to agree on the format to minimize change cycles.  I can discuss live if helpful.

michaeldiamant:
@algochoi https://github.com/algorand/java-algorand-sdk/pull/346/commits/e05aa04eb2fe9ff311bfe3c682ac6a9c8ea546bd applies the feedback - let me know what you think.

A point of context:  Generally, I view moving serialization/encoding into a domain object as a lack of separation of concerns.  Since preferred alternatives (e.g. defining an encoding type class) are limited by the generated code, I'm willing to go down the path for `AppBoxReference`.

algochoi (commented):
Nice, the implementation and the tests LGTM.

Wondering what your thoughts are on naming `BoxQueryEncoding` something more general, like `GoalEncoding`, since this is a goal app-arg convention. I think this would make the class name more future proof if we want to use it in other APIs and provide additional convenience methods for it (e.g. encode to different prefixes like `str` and `int).

michaeldiamant:
> Wondering what your thoughts are on naming BoxQueryEncoding something more general, like GoalEncoding

My 2c is to keep as is though if you'd like to discuss further, I'll happily have a verbal.

Rationale:
* If the encoding scheme is more broadly used, it may no longer be called/considered the _goal_ encoding.
* In the future, there may exist differences in encoding implementation for Boxes vs other concepts applying the encoding scheme.  In my experience, once generalized, it's challenging to break apart concerns.  And it's been easier to start with specificity and generalize later.
* Naming aside, I agree underlying logic _can_ be shared if/when extend usage.

algochoi (approved):
Sounds good, LGTM 📦

algoidurovic (approved):
I'm slightly concerned about users not being aware of this method when calling the API but I don't see a perfect solution to it. Documentation and the locality of having the method available in the class like you're doing here is probably sufficient. Maybe more of an issue in go because of how easy it easy to convert between strings and bytes, making it pretty tempting for the user to call the endpoint with `string(box.name)`. This PR LGTM though, just putting some thoughts out there in case you find them relevant.

PR #347, opened 2022-07-11
Boxes:  Support GetApplicationBoxes

michaeldiamant:
Adds support for GetApplicationBoxes.  

Relates to:
* https://github.com/algorand/go-algorand/pull/4249
* https://github.com/algorand/algorand-sdk-testing/pull/210

### DIFF
--- a/src/main/java/com/algorand/algosdk/util/BoxQueryEncoding.java
+++ b/src/main/java/com/algorand/algosdk/util/BoxQueryEncoding.java
@@ -1,8 +1,8 @@
 package com.algorand.algosdk.util;
 
-import com.algorand.algosdk.transaction.AppBoxReference;
 import com.algorand.algosdk.transaction.Transaction;
 import com.algorand.algosdk.v2.client.model.Box;
+import com.algorand.algosdk.v2.client.model.BoxDescriptor;
 
 /**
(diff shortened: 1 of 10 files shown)


src/main/java/com/algorand/algosdk/util/BoxQueryEncoding.java
```diff
@@ -19,6 +19,10 @@ public static String encodeBox(Box b) {
         return encodeBytes(b.name);
     }
 
+    public static String encodeBoxDescriptor(BoxDescriptor b) {
```
algochoi, line 22: Is this method for the sake of completeness? Would a user need to encode a `BoxDescriptor?

michaeldiamant: @algochoi Primarily, completeness.  Open to discussion - my 2c:
* The SDK can't _safely_ make assumptions about which models will/won't be used.
* For example, consider the following _valid_ (I _think_) scenario using `BoxDescriptor`:  An API user queries all box names for a given app, and then queries for box contents by app + box name.

algochoi: Yes, I think that example makes sense - resolving

algochoi (approved):
LGTM

algochoi (approved):

PR #351, opened 2022-07-28
Enhancement: Add UNKNOWN enum type to HTTP client enums.

winder:
Regenerating client with changes from https://github.com/algorand/generator/pull/40

### DIFF
--- a/src/test/java/com/algorand/algosdk/util/TestEncoder.java
+++ b/src/test/java/com/algorand/algosdk/util/TestEncoder.java
@@ -153,6 +153,8 @@ public class TestEncoder {
         } else {
             assertThat(Transaction.Type.Default).isEqualTo(txn.type);
         }
+
+        TestUtil.serializeDeserializeCheck(txn);
     }
 
     @ParameterizedTest
@@ -170,5 +172,7 @@ public class TestEncoder {
         } else {
             assertThat(Enums.SigType.UNKNOWN).isEqualTo(acct.sigType);
         }
+
+        TestUtil.serializeDeserializeCheck(acct);
     }
 }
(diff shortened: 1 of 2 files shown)


src/test/java/com/algorand/algosdk/util/TestEncoder.java
```diff
@@ -131,4 +137,38 @@ public void testDecodeUint64() {
...
+        }
+    }
+
+    @ParameterizedTest
+    @CsvSource({
+            "sig, sig",
+            "blah, ",
+            "blah, ''"
+    })
+    public void testHttpModelEnum(String type, String expected) throws IOException {
```
michaeldiamant, line 164: @winder Optional - feel welcomed to proceed as is:  I'd feel more comfortable if `testHttpModelEnum` (or a new test) performs round-trip serialization.

The intent is to prove `UNKNOWN` serialization is consistent with the corresponding Jackson annotation configuration.

michaeldiamant (approved):
@winder Looks reasonable - I confirmed re-running _generator_ produces no diff.

src/test/java/com/algorand/algosdk/util/TestEncoder.java
winder: You just reminded me of a utility that does the round trip encoding test for json and msgpack, I updated the test to include that.

It's not perfect though, since this change is specifically making it so that the JSON to object conversion loses information.

michaeldiamant: @winder Thanks for bringing the feedback.  I've taken note of the mentioned utility method - Resolving.

PR #352, opened 2022-08-04
DevTools: adding source map decoder

barnjamin:


src/main/java/com/algorand/algosdk/logic/SourceMap.java
```diff
@@ -0,0 +1,112 @@
...
+
+            this.pcToLine.put(i, lastLine);
+        }
+
+    }
+
+    public Integer getLineForPc(Integer pc) {
+        if(!this.pcToLine.containsKey(pc)){
+            return 0;
+        }
```
michaeldiamant, line 61: @barnjamin What do you think of the following suggestion?

Since we're on Java 7, https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html is unavailable and I don't feel creating a custom type is warranted.  Instead, I propose returning `null` rather than 0 when key does not exist. 

I think the proposed behavior is more inline with other SDKs and less prone to misinterpretation. 
 
```suggestion
```

src/main/java/com/algorand/algosdk/logic/SourceMap.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public ArrayList<Integer> getPcsForLine(Integer line) {
+        if(!this.pcToLine.containsKey(line)){
+            return new ArrayList<Integer>();
+        }
+        return this.lineToPc.get(line);
+    }
+
+    private static class VLQDecoder {
```
michaeldiamant, line 72: @barnjamin Checking:  Is there a citation that be added inline to help other maintainers understand where the VLQ decoding implementation came from?

I'm assuming (perhaps incorrectly) that it's based on one of the other SDKs.

src/main/java/com/algorand/algosdk/logic/SourceMap.java
michaeldiamant: Optionally - If you take the _suggestion_, I'd consider adding a Javadoc comment highlighting the return value semantics.

src/main/java/com/algorand/algosdk/logic/SourceMap.java
barnjamin: yes i took it from the go sdk, i can add some documentation to provide context

michaeldiamant: @barnjamin Thanks for the note - looks reasonable, resolving.

michaeldiamant (approved):

PR #354, opened 2022-08-10
Merge develop branch to boxes feature branch

algochoi:


ahangsu (approved):
`Makefile` conflict resolving seems making sense to me.

Makefile
```diff
@@ -1,8 +1,8 @@
...
 
 integration:
-	mvn test -Dcucumber.filter.tags="@algod or @assets or @auction or @kmd or @send or @send.keyregtxn or @indexer or @rekey or @applications.verified or @applications or @applications.boxes or @compile or @dryrun or @indexer.applications or @indexer.231 or @abi or @c2c"
+	mvn test -Dcucumber.filter.tags="@algod or @assets or @auction or @kmd or @send or @send.keyregtxn or @indexer or @rekey_v1 or @applications.verified or @applications or @applications.boxes or @compile or @dryrun or @indexer.applications or @indexer.231 or @abi or @c2c or @compile.sourcemap"
```
michaeldiamant, line 5: @algochoi Sanity checking:  Does the PR deliberately remove the following `develop` addition?

https://github.com/algorand/java-algorand-sdk/blob/5db404e6fb310cf9c079693e0c09eb7a9863395c/Makefile#L6

ahangsu: I haven't seen this line before in java-sdk, seems introduced in sdk testing improvement PR, what is this for?

my bad for missing this line.

algochoi: The integration tests seem to pass anyway - was this previously needed?

ahangsu: I did search around, and found this: https://stackoverflow.com/questions/1873995/run-a-single-test-method-with-maven, and wondered if this line have any influence on integration test, for the test being pointed at is empty.

michaeldiamant: @algochoi @ahangsu Let me know if the note helps clarify?

* `RunCucumberIntegrationTest` is the entry point for integration Cucumber test execution.
  * Admittedly, there's a bit of wiring between JUnit + Cucumber that I've not fully researched.  I suspect there's duplication `RunCucumberUnitTest`, but let's save that for another time. 
* Consequently, when executing `make integration`, the _only_ test Maven needs to target is `RunCucumberIntegrationTest`. 
* Without specifying the filter, `make integration` re-runs all non-Cucumber unit tests during integration test execution.  
  * So, the flag is not strictly necessary.  However, I think it's helpful to limit console output in order to disambiguate test output + speed up execution.
  * No Cucumber unit tests are run because no corresponding tags are provided in `-Dcucumber.filter.tags`.

algochoi: @michaeldiamant thanks for the explanation - I'll put the line back in 👍

michaeldiamant: @algochoi Looks good as of https://github.com/algorand/java-algorand-sdk/pull/354/commits/9529ec1d4c154ac88dae026f1e68eac2b2afb22b.  Resolving.

michaeldiamant (approved):

Makefile
ahangsu: Tho its a bit late but, thanks for explaining!

PR #360, opened 2022-08-16
StateProofs: Add State Proof support.

winder:
Add support for new state proof endpoints and transaction deserialization.


src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -162,6 +162,16 @@ public class Transaction implements Serializable {
...
 
+    /* state proof fields */
+    @JsonProperty("sptype")
+    public BigInteger stateProofType = null;
+
+    @JsonProperty("sp")
+    public Map<String,Object> stateProof = null;
+
+    @JsonProperty("spmsg")
+    public Map<String,Object> stateProofMessage = null;
```
winder, line 173: These types were really complicated, do they need to be added?

winder, line 173: These types were really complicated, do they need to be added?

src/main/java/com/algorand/algosdk/v2/client/model/IndexerStateProofMessage.java
```diff
@@ -0,0 +1,68 @@
...
+    public String blockHeadersCommitment() {
+        return Encoder.encodeToBase64(this.blockHeadersCommitment);
+    }
+    public byte[] blockHeadersCommitment;
+
+    /**
+     * (f)
+     */
+    @JsonProperty("first-attested-round")
+    public java.math.BigInteger firstAttestedRound;
```
shiqizng, line 27: are these code generated? I think long type should be sufficient.

shiqizng (commented):
this looks good. Using BigInteger type seems odd to me.

README.md
```diff
@@ -229,6 +229,14 @@ mvn clean site -P github,default  # for javadoc
 mvn clean deploy -P release,default
 ```
 
+# Testing
+
+Many cross-SDK tests are defined in [algorand-sdk-testing](https://github.com/algorand/algorand-sdk-testing/). Some are integration tests wit additional dependencies. These dependencies are containerized in a docker file, which can be executed with `make docker-test`.
```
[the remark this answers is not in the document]
winder:
```suggestion
Many cross-SDK tests are defined in [algorand-sdk-testing](https://github.com/algorand/algorand-sdk-testing/). Some are integration tests with additional dependencies. These dependencies are containerized in a docker file, which can be executed with `make docker-test`.
```

src/main/java/com/algorand/algosdk/v2/client/model/IndexerStateProofMessage.java
winder: This one is generated code. I think it means the type was tagged with `uint64`, so the generator uses a BigInteger.

shiqizng: I see.

shiqizng (approved):

algoidan:
think we should remove the old `client/v2/algod/GetProof.java` file since the endpoind was renamed

src/main/java/com/algorand/algosdk/v2/client/common/AlgodClient.java
```diff
@@ -140,12 +142,12 @@ public GetBlock GetBlock(Long round) {
     }
 
     /**
-     * Get a Merkle proof for a transaction in a block.
+     * Get a proof for a transaction in a block.
      * /v2/blocks/{round}/transactions/{txid}/proof
      */
-    public GetProof GetProof(Long round,
+    public GetTransactionProof GetTransactionProof(Long round,
```
algoidan, line 148: The API now supports one more parameter which is the hash type
should we add a parameter to this function?

winder: That is an optional query parameter, the arguments here only include required path parameters.

algoidan: You are right. it is indeed an optional param. However, this API is a  significant part of "how to build a light client" and how to bridge tokens, and currently, its default value does not match the light client verification so it would make it usable for the light client use case

Eric-Warehime: It's  _not_ unusable--required parameters go in the initial constructor, and optional parameters can be specified in a builder-like pattern like so
```java
TransactionProofResponse resp = client.GetTransactionProof(123456, "abc123").hashtype(Enums.HashType("sha256_512")).execute();
```

algoidan:
should we apply changes to the header model as well?
https://github.com/algorand/java-algorand-sdk/blob/d9b2699843e237ebfa4303d8a744670a827083ea/src/main/java/com/algorand/algosdk/algod/client/model/Block.java#L75

algoidan (approved):

algoidan (approved):

PR #367, opened 2022-08-19
Enhancement: Deprecating use of langspec

ahangsu:


src/main/java/com/algorand/algosdk/crypto/LogicsigSignature.java
```diff
@@ -48,14 +48,17 @@ public LogicsigSignature(
...
         this.args = args;
-        boolean verified = false;
-        try {
-            verified = Logic.checkProgram(this.logic, this.args);
-        } catch (IOException ex) {
-            throw new IllegalArgumentException("invalid program", ex);
-        }
+        Logic.sanityCheckProgram(this.logic);
+
+        // Deprecated:
```
michaeldiamant, line 53: @ahangsu Do you wish to retain the commented out logic or is it temporary?

Asking now because I imagine the same diff will appear in other SDKs.  Open to discussion, but from my perspective, the internal logic (like the highlighted example) can be deleted along with associated tests.

src/main/java/com/algorand/algosdk/logic/Logic.java
```diff
@@ -158,6 +220,7 @@ public static VarintResult getUVarint(byte [] buffer, int bufferOffset) {
      * @return
      * @throws IOException
      */
+    @Deprecated
```
michaeldiamant, line 223: @ahangsu What's your thinking on the following?

Since `checkProgram` and `readProgram` appear to be the most centrally exposed (public) methods, I'm thinking it'd be helpful to add a bit more context via a javadoc `@deprecated` tag (because Java 7 `@Deprecated` lacks fields) on _why_ it's deprecated + stand-in replacement methods.

I can be convinced to leave as is.  I'm hoping there's a lightweight 1-2 sentences that can be added in here rather than something more onerous.

src/main/java/com/algorand/algosdk/crypto/LogicsigSignature.java
ahangsu: Oh, this is just temporary, for the sake of a visual reference on some testcases, these will be deleted later on.

src/main/java/com/algorand/algosdk/logic/Logic.java
ahangsu: yes, the `@deprecated` tag in `java-docs` totally makes sense, as a detained explanation for the decision on deprecation, and this can also be extended to other SDKs.

ahangsu: just added some example message under `@deprecation` tag.

src/main/java/com/algorand/algosdk/crypto/LogicsigSignature.java
```diff
@@ -48,14 +48,10 @@ public LogicsigSignature(
...
-        try {
-            verified = Logic.checkProgram(this.logic, this.args);
-        } catch (IOException ex) {
-            throw new IllegalArgumentException("invalid program", ex);
-        }
 
-        assert verified;
+        // NOTE: Logic.checkProgram has been deprecated, use Logic.sanityCheckProgram instead
+        //       we no longer read from langspec.json, we do not need to catch IOException error
+        Logic.sanityCheckProgram(this.logic);
```
michaeldiamant, line 54: @ahangsu Minor and optional:  For replacements _not_ exposed publicly, I don't think an inline citation is required.  If you prefer it, feel welcomed to keep as is.


```suggestion
        Logic.sanityCheckProgram(this.logic);
```

src/main/java/com/algorand/algosdk/crypto/LogicsigSignature.java
```diff
@@ -124,11 +120,9 @@ public boolean verify(Address singleSigner) throws NoSuchAlgorithmException {
...
         }
 
-        try {
-            Logic.checkProgram(this.logic, this.args);
-        } catch (Exception ex) {
-            return false;
-        }
+        // NOTE: Logic.checkProgram has been deprecated, use Logic.sanityCheckProgram instead
+        //       we no longer read from langspec.json, we do not need to catch IOException error
+        Logic.sanityCheckProgram(this.logic);
```
michaeldiamant, line 125: @ahangsu Minor and optional:  For replacements _not_ exposed publicly, I don't think an inline citation is required.  If you prefer it, feel welcomed to keep as is.

```suggestion
        Logic.sanityCheckProgram(this.logic);
```

src/main/java/com/algorand/algosdk/logic/Logic.java
```diff
@@ -147,28 +164,56 @@ public static VarintResult getUVarint(byte [] buffer, int bufferOffset) {
         return new VarintResult();
     }
 
+    public static void sanityCheckProgram(final byte[] program) {
```
michaeldiamant, line 167: @ahangsu Leaving a visual reminder to revisit once we agree on `sanityCheckProgram` internals:  The method should provide javadoc to summarize usage to API users.

.test-env
```diff
@@ -1,6 +1,7 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+# TODO change back to master when sdk-testing is merged
+SDK_TESTING_BRANCH="deprecate-langspec"
```
ahangsu, line 4: after algorand-sdk-testing is merged, do
```suggestion
SDK_TESTING_BRANCH="master"
```

src/main/java/com/algorand/algosdk/crypto/LogicsigSignature.java
```diff
@@ -39,6 +40,60 @@ public class LogicsigSignature {
     @JsonProperty("msig")
     public MultisigSignature msig;
 
+
+    private static boolean isAsciiPrintable(final byte symbol) {
+        char symbolChar = (char) (symbol & 0xFF);
```
jasonpaulos, line 45: Is the masking necessary here? Would `(char) symbol` be sufficient?

michaeldiamant (approved):

src/main/java/com/algorand/algosdk/crypto/LogicsigSignature.java
ahangsu: yea, `(char) symbol` is just sufficient

ahangsu: actually, we don't even have to do the byte to char conversion, I am taking that out.

jasonpaulos (approved):

PR #388, opened 2022-09-12
Deprecation: Add deprecation tags to v1 `algod` API

algochoi:
This PR adds deprecation tags to the Javadocs to v1 algod APIs and related models/utilities.

Related issue here: https://github.com/algorand/algorand-sdk-testing/issues/218

### DIFF
--- a/src/main/java/com/algorand/algosdk/algod/client/JSON.java
+++ b/src/main/java/com/algorand/algosdk/algod/client/JSON.java
@@ -32,6 +32,7 @@ import java.text.ParsePosition;
 import java.util.Date;
 import java.util.Map;
 
+@Deprecated
 public class JSON {
     private Gson gson;
     private boolean isLenientOnJson = false;
--- a/src/main/java/com/algorand/algosdk/algod/client/StringUtil.java
+++ b/src/main/java/com/algorand/algosdk/algod/client/StringUtil.java
@@ -13,7 +13,7 @@
 
 package com.algorand.algosdk.algod.client;
 
-
+@Deprecated
 public class StringUtil {
   /**
    * Check if the given array contains the given value (with case-insensitive comparison).
--- a/src/main/java/com/algorand/algosdk/abi/Contract.java
+++ b/src/main/java/com/algorand/algosdk/abi/Contract.java
@@ -1,6 +1,5 @@
 package com.algorand.algosdk.abi;
 
-import com.algorand.algosdk.algod.client.StringUtil;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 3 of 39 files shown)


michaeldiamant (approved):
@algochoi Matches my expectations of what we agreed to deprecate.  Prefer a 2nd reviewer approves prior to merge.

src/main/java/com/algorand/algosdk/algod/client/StringUtil.java
```diff
@@ -13,7 +13,7 @@
 
 package com.algorand.algosdk.algod.client;
 
-
+@Deprecated
 public class StringUtil {
```
ahangsu, line 17: nit: I have some different thoughts on it. If you look at this class's usage, then it gets into ABI implementation of some string concatenation. Shell we mark this one as deprecated, and *also take care of the ABI case*?

src/main/java/com/algorand/algosdk/algod/client/JSON.java
```diff
@@ -32,6 +32,7 @@
 import java.util.Date;
 import java.util.Map;
 
+@Deprecated
 public class JSON {
```
ahangsu, line 36: I noticed there are some public static class declaration inside `JSON`, do we want to mark them as deprecated?

src/main/java/com/algorand/algosdk/algod/client/StringUtil.java
algochoi: Good observation - I think we should use `org.apache.commons.lang3.StringUtils` since we have already added that dependency in `src/main/java/com/algorand/algosdk/v2/client/Utils.java`. 

Otherwise, we *could* use the `join` in `src/main/java/com/algorand/algosdk/kmd/client/StringUtil.java`, since `kmd` is not being deprecated. I think using `commons.lang3.StringUtils` is a better option though, and if you agree, I can make the change in this PR in the ABI classes.

src/main/java/com/algorand/algosdk/algod/client/JSON.java
algochoi: I think externally, it should still be marked as deprecated (e.g. if you do something like `JSON.static_class()`, the annotation will appear on the first `JSON`). Admittedly, I only tested this on the latest `javac` version.

winder:
How are you determining the overlap between Kmd and Algod v1 functionality? The most convenient reference I have for KMD is the README example, it imports the following:
```
import com.algorand.algosdk.kmd.client.ApiException;
import com.algorand.algosdk.kmd.client.KmdClient;
import com.algorand.algosdk.kmd.client.api.KmdApi;
import com.algorand.algosdk.kmd.client.model.*;
```

src/main/java/com/algorand/algosdk/algod/client/StringUtil.java
ahangsu: Great - If `org.apache.commons.lang3.StringUtils` is available, then we should change the dependencies in ABI classes, and I think we can include these changes inside this PR, no need to open yet another PR.

src/main/java/com/algorand/algosdk/algod/client/JSON.java
ahangsu: got it, then I resolve it.

winder (approved):

PR #390, opened 2022-09-13
enhancement: update block model

shiqizng:
Updating block model to include ParticipationUpdates.  cucumber test update [PR](https://github.com/algorand/algorand-sdk-testing/pull/234).

### DIFF
--- a/src/main/java/com/algorand/algosdk/v2/client/model/Block.java
+++ b/src/main/java/com/algorand/algosdk/v2/client/model/Block.java
@@ -34,6 +34,12 @@ public class Block extends PathResponse {
     public String genesisId;
 
     /**
+     * Participation account data that needs to be checked/acted on by the network.
+     */
+    @JsonProperty("participation-updates")
+    public ParticipationUpdates participationUpdates;
+
+    /**
      * (prev) Previous block hash.
      */
     @JsonProperty("previous-block-hash")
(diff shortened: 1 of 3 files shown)


src/main/java/com/algorand/algosdk/v2/client/model/Block.java
```diff
@@ -33,6 +33,12 @@ public String genesisHash() {
     @JsonProperty("genesis-id")
     public String genesisId;
 
+    /**
+     * Participation account data that needs to be checked/acted on by the network.
+     */
+    @JsonProperty("participation-updates")
+    public ParticipationUpdates participationUpdates;
```
jasonpaulos, line 40: Are these code changes generated, or hand written?

jasonpaulos (commented):
Seems to make sense, but I suppose we need to wait until https://github.com/algorand/algorand-sdk-testing/pull/234 is merged to have CI confirm

src/main/java/com/algorand/algosdk/v2/client/model/Block.java
shiqizng: generated.

shiqizng:
sdk-testing updated

jasonpaulos (approved):
Thanks, looks good

winder (approved):

PR #401, opened 2022-09-22
Re-run code generator

michaeldiamant:
Addresses https://github.com/algorand/java-algorand-sdk/pull/345#discussion_r976911364 by re-running _generator_ using branches matching `.env` in https://github.com/algorand/algorand-sdk-testing/pull/192.


ahangsu (approved):

src/main/java/com/algorand/algosdk/v2/client/model/BoxesResponse.java
```diff
@@ -21,6 +21,14 @@ public class BoxesResponse extends PathResponse {
     @JsonProperty("boxes")
     public List<BoxDescriptor> boxes = new ArrayList<BoxDescriptor>();
 
+    /**
+     * Base64 encoded final box name result. Used for pagination, when making another
+     * request provide this token with the next parameter and prepend with "b64:" if
+     * keeping the provided encoding.
```
jasonpaulos, line 27: I noticed this description is not accurate (we don't require the user to prefix the next token with `b64:`), so I made an indexer PR to correct it: https://github.com/algorand/indexer/pull/1240

@michaeldiamant I'll defer to your judgement about how you wish to proceed. Acceptable solutions to me include merging this as-is and following up with a later PR, or holding this until the indexer PR is adopted.

jasonpaulos (commented):
I noticed one minor issue, but otherwise this looks good

src/main/java/com/algorand/algosdk/v2/client/model/BoxesResponse.java
michaeldiamant: @jasonpaulos Good catch - I opted to refresh once more because the PR is still open via https://github.com/algorand/java-algorand-sdk/pull/401/commits/634d58dca5fd8edae9476adb6457f34e0ef3fa5b.  Can you have a look?

Aside - I had already also amended our merge checklist to require ourselves to regenerate sources prior to merging SDK PRs.

jasonpaulos: Thanks for the quick incorporation, I'll resolve this

jasonpaulos (approved):

PR #408, opened 2022-09-27
Rework MethodCallParams to be backwards compatible with Boxes

michaeldiamant:
Reworks `MethodCallParams` to be backwards compatible with Boxes by making a protected constructor available to the builder.  Additionally, deprecates the public constructor to clearly warn about usage.

### DIFF
--- a/src/main/java/com/algorand/algosdk/transaction/MethodCallParams.java
+++ b/src/main/java/com/algorand/algosdk/transaction/MethodCallParams.java
@@ -36,7 +36,7 @@ public class MethodCallParams {
     public final List<Address> foreignAccounts;
     public final List<Long> foreignAssets;
     public final List<Long> foreignApps;
-    public List<AppBoxReference> boxReferences;
+    public final List<AppBoxReference> boxReferences;
 
     public final TEALProgram approvalProgram, clearProgram;
     public final StateSchema globalStateSchema, localStateSchema;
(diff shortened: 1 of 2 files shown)


michaeldiamant:
Additional reviewer context:  I originally aimed to provide a `private` `MethodCallParams` constructor.  However, the builder lives in a different package.  So, it appears a `protected` constructor is the best available alternative without (larger and out-of-scope) package moves.

src/main/java/com/algorand/algosdk/transaction/MethodCallParams.java
```diff
@@ -35,7 +36,7 @@ public class MethodCallParams {
     public final List<Address> foreignAccounts;
     public final List<Long> foreignAssets;
     public final List<Long> foreignApps;
-    public final List<AppBoxReference> boxReferences;
+    public List<AppBoxReference> boxReferences;
```
jasonpaulos, line 39: I believe this field is meant to stay `final`?

jasonpaulos (commented):
Generally looks good, there's just one issue I found

src/main/java/com/algorand/algosdk/transaction/MethodCallParams.java
michaeldiamant: @jasonpaulos Thanks for catching and sorry for missing on my end.  It's addressed in https://github.com/algorand/java-algorand-sdk/pull/408/commits/d79e9fcd7f25bd82fbb6ef53fab0fadca7faaaf3.  Can you resolve if it looks good?

ahangsu (approved):
The change looks good to me at this point, so I am leaving an approval.

jasonpaulos (approved):
Looks good now, thanks

PR #413, opened 2022-10-05
REST API: Add algod block hash endpoint, add indexer block header-only param.

winder:
Add tests to generated code. See commits for the generated code vs the tests.

Tests are passing locally, CI is pending https://github.com/algorand/algorand-sdk-testing/pull/245


tzaffi (commented):
seems good to me. We should base the changes in the harness script off the definitive version of it in https://github.com/algorand/go-algorand-sdk/pull/421

tzaffi (approved):
LGTM

PR #425, opened 2022-11-15
Enhancement: Migrate v1 algod dependencies to v2 in cucumber tests

ahangsu:
Migrates v1 algod usage in cucumber integration tests to use v2. Also deletes some unused steps relating to v1 APIs.

SDK Testing PR: https://github.com/algorand/algorand-sdk-testing/pull/249


src/test/java/com/algorand/algosdk/integration/Stepdefs.java
```diff
@@ -352,7 +260,7 @@ public void mn_for_sk(String mn) throws GeneralSecurityException{
     }
 
     @Given("default V2 key registration transaction {string}")
-    public void  default_v2_key_registration_transaction(String type) throws NoSuchAlgorithmException, JsonProcessingException, IOException{
+    public void  default_v2_key_registration_transaction(String type) {
```
algochoi, line 263:
``` suggestion
public void default_v2_key_registration_transaction(String type) {
```

src/test/java/com/algorand/algosdk/integration/Stepdefs.java
```diff
@@ -989,14 +832,9 @@ public void checkMicro(long ma) {
...
 
-    @Then("I get account information")
-    public void accInfo() throws ApiException {
-        acl.accountInformation(addresses.get(0));
-    }
-
     @Then("I can get account information")
-    public void newAccInfo() throws ApiException, NoSuchAlgorithmException, com.algorand.algosdk.kmd.client.ApiException {
-        acl.accountInformation(pk.encodeAsString());
+    public void newAccInfo() throws Exception {
```
algochoi, line 836: Is it okay to use the general `Exception` here instead of specific exception instances?

src/test/java/com/algorand/algosdk/integration/Stepdefs.java
```diff
@@ -1033,21 +871,25 @@ public void default_asset_creation_transaction_with_total_issuance(Integer asset
     }
 
     @When("I get the asset info")
-    public void i_get_the_asset_info() throws ApiException {
-        this.queriedParams = acl.assetInformation(this.assetID);
+    public void i_get_the_asset_info() throws Exception {
+        this.queriedParams = aclv2.GetAssetByID(this.assetID.longValue()).execute().body();
+    }
+
+    static private String null_to_empty(String str) {
```
algochoi, line 878: :+1:

src/test/java/com/algorand/algosdk/integration/Stepdefs.java
```diff
@@ -1077,127 +919,118 @@ public void i_create_an_asset_destroy_transaction() throws NoSuchAlgorithmExcept
...
     @Then("I should be unable to get the asset info")
-    public void i_should_be_unable_to_get_the_asset_info() {
-        boolean exists = true;
-        try {
-            this.i_get_the_asset_info();
-        } catch (ApiException e) {
-            exists = false;
-        }
-        assertThat(exists).isFalse();
+    public void i_should_be_unable_to_get_the_asset_info() throws Exception {
```
algochoi, line 922: Same question about catching `Exception` above.

algochoi (commented):
Nice work here, thanks for the efforts in updating v1 algod APIs - minor comments/questions below

src/test/java/com/algorand/algosdk/integration/Applications.java
```diff
@@ -15,16 +15,13 @@
 import io.cucumber.java.en.Then;
 import org.apache.commons.lang3.StringUtils;
 import org.assertj.core.api.Assertions;
-import org.assertj.core.util.Lists;
 import org.bouncycastle.util.Strings;
```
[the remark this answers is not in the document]
ahangsu: good substitution

src/test/java/com/algorand/algosdk/integration/Stepdefs.java
ahangsu: I think it is because of the `execute()` method in `aclv2` client, that throws a general `Exception`.

algochoi: Ah interesting, it looks like it's due to some code here: algorand/java-algorand-sdk/src/main/java/com/algorand/algosdk/v2/client/common/Client.java

It looks like the client dependency uses an [`IOException`](https://square.github.io/okhttp/2.x/okhttp/com/squareup/okhttp/Call.html#execute--) but we previously declared it to be an `Exception` instead. 

I think this is outside the scope of this PR anyway, so I'll close this.

src/test/java/com/algorand/algosdk/integration/Stepdefs.java
algochoi: Addressed in above comment

src/test/java/com/algorand/algosdk/integration/Applications.java
```diff
@@ -142,18 +137,19 @@ public void sendTransactionWithTransientAccountAndCheckForError(String error) th
 
         // And save the txId for later
         this.txId = rPost.body().txId;
+        base.txid = this.txId;
```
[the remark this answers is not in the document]
ahangsu: There is a difference - the wait for transaction need the transaction id, and the wait function is in stepdef.java. We need to pass the `txId` to `base` such that the function waits for right transaction.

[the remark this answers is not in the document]
algochoi: makes sense, thanks!

algochoi (approved):
Updated tests look good and it looks like it ran correctly in Circle CI - LGTM

Good to merge after changing the SDK testing branch

src/test/java/com/algorand/algosdk/integration/Applications.java
```diff
@@ -142,18 +136,19 @@ public void sendTransactionWithTransientAccountAndCheckForError(String error) th
...
-        Utils.waitForConfirmation(clients.v2Client, txId, 1);
+        Utils.waitForConfirmation(base.aclv2, base.txid, 1);
     }
 
     // TODO: Use V2 Pending Transaction endpoint when it is available.
     //       The initial implementation hacks into the v1 endpoint to manually extract the new data.
     @Given("I remember the new application ID.")
     public void rememberTheNewApplicatoinId() throws Exception {
-        PendingTransactionResponse r = clients.v2Client.PendingTransactionInformation(txId).execute().body();
+        PendingTransactionResponse r = base.aclv2.PendingTransactionInformation(txId).execute().body();
```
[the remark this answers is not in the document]
ahangsu: good point, thanks for pointing out

PR #431, opened 2022-11-28
Bug-Fix: parsing type strings for tuples containing static arrays of tuples

ahangsu:
Fix https://github.com/algorand/avm-abi/issues/12

### DIFF
--- a/src/main/java/com/algorand/algosdk/abi/ABIType.java
+++ b/src/main/java/com/algorand/algosdk/abi/ABIType.java
@@ -103,7 +103,7 @@ public abstract class ABIType {
             return new ArrayList<>();
 
         if (str.startsWith(",") || str.endsWith(","))
-            throw new IllegalArgumentException("parsing error: tuple content should not start with comma");
+            throw new IllegalArgumentException("parsing error: tuple content should not start or end with comma");
 
         if (str.contains(",,"))
             throw new IllegalArgumentException("parsing error: tuple content should not have consecutive commas");
--- a/src/main/java/com/algorand/algosdk/abi/ABIType.java
+++ b/src/main/java/com/algorand/algosdk/abi/ABIType.java
@@ -119,6 +119,8 @@ public abstract class ABIType {
                     throw new IllegalArgumentException("parsing error: tuple parentheses are not balanced: " + str);
                 int leftParenIndex = parenStack.pop();
                 if (parenStack.isEmpty()) {
+                    // iterate through the byte str, include all the bytes after closing round bracket, for array indicator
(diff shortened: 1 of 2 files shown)


src/main/java/com/algorand/algosdk/abi/ABIType.java
```diff
@@ -118,8 +118,13 @@ else if (str.charAt(i) == ')') {
                 if (parenStack.isEmpty())
```
bbroder-algo, line 118: similar error in the exception above in 106

src/test/java/com/algorand/algosdk/abi/TestTypes.java
```diff
@@ -238,6 +238,22 @@ public void TestTypeFromStringValid() {
                         )
                 )
         );
+        assertThat(ABIType.valueOf("(uint32,(uint64,bool)[10],byte)")).isEqualTo(
```
bbroder-algo, line 241: maybe an additional one that puts the static array at the end so it tests the end-of-string condition in addition to the comma condition.

src/main/java/com/algorand/algosdk/abi/ABIType.java
ahangsu: 👍 good find

src/test/java/com/algorand/algosdk/abi/TestTypes.java
ahangsu: got it, I will mimic the cases in avm-abi one

algochoi (approved):
Thanks for the fix + tests look good!

src/main/java/com/algorand/algosdk/abi/ABIType.java
```diff
@@ -118,8 +118,13 @@ else if (str.charAt(i) == ')') {
...
                     throw new IllegalArgumentException("parsing error: tuple parentheses are not balanced: " + str);
                 int leftParenIndex = parenStack.pop();
-                if (parenStack.isEmpty())
+                if (parenStack.isEmpty()) {
+                    int forwardIndex = i + 1;
```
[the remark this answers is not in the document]
ahangsu: 👍 will do

michaeldiamant (approved):
@ahangsu Thanks for supplying a fix - As intended, I confirmed:
* Added tests fail prior to the PR.
* Implementation mirrors other SDKs.

Out-of-scope passing remark:  I can imagine building a property-based test as a way to build more confidence in test case generation.  Though it's a big enough effort that I don't think it's worth pursuing in-scope to the story.

Issue #389, opened 2022-09-12
Breaking changes in v2.0.0


michaeldiamant:
Itemizes functionality that will be removed in the upcoming major release.

Task list for algorand-sdk-testing stories:
- [x] https://github.com/algorand/algorand-sdk-testing/issues/201 - https://github.com/algorand/java-algorand-sdk/pull/445
- [x] https://github.com/algorand/algorand-sdk-testing/issues/202 - https://github.com/algorand/java-algorand-sdk/pull/441
- [x] https://github.com/algorand/algorand-sdk-testing/issues/223 - https://github.com/algorand/java-algorand-sdk/pull/442
- [x] https://github.com/algorand/algorand-sdk-testing/issues/254 - https://github.com/algorand/java-algorand-sdk/pull/442
- [x] https://github.com/algorand/algorand-sdk-testing/issues/255 - https://github.com/algorand/java-algorand-sdk/pull/443
- [x] Remove the deprecated `MethodCallParams` public constructor.  Deprecation added in https://github.com/algorand/java-algorand-sdk/pull/408. - https://github.com/algorand/java-algorand-sdk/pull/443
- [x] https://github.com/algorand/generator/issues/46 - https://github.com/algorand/generator/pull/50

PR #441, opened 2022-12-12
Enhancement: Sourcemap `mapping` field deprecation

ahangsu:
Implements https://github.com/algorand/algorand-sdk-testing/issues/202 here

### DIFF
--- a/.test-env
+++ b/.test-env
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="revert-sandbox"
+SDK_TESTING_BRANCH="master"
 SDK_TESTING_HARNESS="test-harness"
 
 INSTALL_ONLY=0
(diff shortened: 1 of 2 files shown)


.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="revert-sandbox"
```
ahangsu, line 3:
```suggestion
SDK_TESTING_BRANCH="master"
```

algochoi (approved):
I think this looks good - since the test itself is not a breaking change, we could also target `master` for this PR right?

ahangsu:
I think for now all the testing branches are targeting master?

algochoi:
Oh I mean this PR itself (currently targeting [feature/v2.0.0-break](https://github.com/algorand/java-algorand-sdk/tree/feature/v2.0.0-break))

ahangsu:
I think it is part of https://github.com/algorand/java-algorand-sdk/issues/389, so I'm pointing to feature branch for now.

ahangsu:
After more thoughts, I think you are right. If this PR does not go to master, then the current change of testcase might fail.

PR #450, opened 2022-12-14
Remove deprecated methods in Transaction.java

ahangsu:
Most of the methods in `Transaction.java` marked as deprecated, are created around 2 years ago. Maybe it is time to remove them at this moment.


michaeldiamant:
For reader reference - https://github.com/algorand/java-algorand-sdk/pull/105 deprecated the `Transaction` constructors + factory methods.

src/main/java/com/algorand/algosdk/account/Account.java
```diff
@@ -219,35 +219,6 @@ public SignedBid signBid(Bid bid) throws NoSuchAlgorithmException {
         }
     }
```
michaeldiamant, line 221: @ahangsu @winder @algochoi @jasonpaulos What are your thoughts on extending the PR to either remove _or_ mark private the _non_-deprecated public constructor? 
https://github.com/algorand/java-algorand-sdk/blob/310f3badb348b51392d80414fc4de57da19dc4d6/src/main/java/com/algorand/algosdk/transaction/Transaction.java#L336

* Usages are:
  * The private `Transaction` constructor invoked by JSON deserialization.
  * Several deprecated factory methods (e.g. `Transaction.createPaymentTransaction`).  I think these factory methods should be removed. 
* Admittedly, the constructor was _not_ deprecated in advance of v2.0.0.  However, the builder pattern has been in use for ~2 years.  
* Open to feedback - I propose:  
  * Remove the constructor.  I am willing to settle for marking private.
  * I _think_ we must retain the no-arg constructor (`public Transaction()`) for Jackson.  Though I suggest testing the assumption.

winder: Good suggestion. +1 to marking as private or removing.

It's true that the no-arg constructor is required for Jackson, but I don't know if it needs to be public. This requirement comes from the `NON_DEFAULT` serialization pattern that we have in order to comply with the messagepack encoding strategy. You can see the documented behavior here: https://github.com/FasterXML/jackson-databind/blob/92da0c813979936f792f99a779ae861d1ee3839e/src/main/java/com/fasterxml/jackson/databind/annotation/JsonSerialize.java#L210

ahangsu: > It's true that the no-arg constructor is required for Jackson, but I don't know if it needs to be public.

The public no-arg constructor is need by transaction builder and signed transaction... I don't think there is an easy way to get around this one.

> Remove the constructor. I am willing to settle for marking private.

I think we want to keep the constructor private, for we still need it to initialize class. Pushed something in [5dd8560](https://github.com/algorand/java-algorand-sdk/pull/450/commits/5dd8560863f5319693e62f511081dc0212194d17), to mark constructor private.

michaeldiamant: @ahangsu Got it - Makes sense that further removal of constructor access is more involved.  I think we've got a good improvement here.

Before resolving - There's several unused private constructors.  Can you review the following git patch to see if you agree with it?

[diff.txt](https://github.com/algorand/java-algorand-sdk/files/10239491/diff.txt)

ahangsu: I am fine with it, I think these 2 constructors should be used by previous deprecated methods, should be okay to remove.

src/main/java/com/algorand/algosdk/transaction/Transaction.java
```diff
@@ -175,367 +175,6 @@ public class Transaction implements Serializable {
     @JsonProperty("spmsg")
```
michaeldiamant, line 175: @ahangsu Minor:  Visual reminder to add a CHANGELOG once we get closer to the finish line.

src/main/java/com/algorand/algosdk/account/Account.java
michaeldiamant: @ahangsu Resolving - Thanks for taking in multiple rounds of feedback here.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Remove `TxGroup.assignGroupID(Transaction[] txns, Address address)` in favor
   of `TxGroup.assignGroupID(Address address, Transaction ...txns)`.
 * Remove `Account.transactionWithSuggestedFeePerByte` in favor of `Account.setFeeByFeePerByte`.
+* Remove deprecated methods in `Transaction.java`, mark public `Transaction` constructor as hidden in favor of `builder.transaction`.
```
michaeldiamant, line 8: @ahangsu Optional - Where possible, I prefer a fully qualified package name to minimize confusion.  Though I don't feel strongly.

```suggestion
* Remove deprecated methods in `Transaction.java`, mark public `Transaction` constructor as hidden in favor of `com.algorand.algosdk.builder.transaction`.
```

michaeldiamant (approved):

PR #452, opened 2022-12-15
Remove deprecated `setFee` and `setLease` in `Transaction`

ahangsu:
Remove remaining deprecated methods in `Transaction`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   of `TxGroup.assignGroupID(Address address, Transaction ...txns)`.
 * Remove `Account.transactionWithSuggestedFeePerByte` in favor of `Account.setFeeByFeePerByte`.
 * Remove deprecated methods in `Transaction.java`, mark public `Transaction` constructor as hidden in favor of `builder.transaction`.
+* Remove deprecated `Transaction.setFee` and `Transaction.setLease` methods
 
 # 1.22.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
   of `TxGroup.assignGroupID(Address address, Transaction ...txns)`.
 * Remove `Account.transactionWithSuggestedFeePerByte` in favor of `Account.setFeeByFeePerByte`.
 * Remove deprecated methods in `Transaction.java`, mark public `Transaction` constructor as hidden in favor of `com.algorand.algosdk.builder.transaction`.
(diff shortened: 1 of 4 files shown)


src/main/java/com/algorand/algosdk/account/Account.java
```diff
@@ -247,7 +247,7 @@ static public void setFeeByFeePerByte(Transaction tx, BigInteger suggestedFeePer
         if (fee.compareTo(Account.MIN_TX_FEE_UALGOS) < 0) {
             fee = Account.MIN_TX_FEE_UALGOS;
         }
-        tx.setFee(fee);
+        tx.fee = fee;
```
algochoi, line 250: nit: a little strange that `tx.fee` is initialized twice in line 243 and here

algochoi (commented):
Nice work, do you know if `setNote` should also be deprecated in favor of transaction builder?

src/main/java/com/algorand/algosdk/account/Account.java
ahangsu: I am unaware of the previous initialization, but this function do not return early, so I am fine with just removing it.

ahangsu: ok it is interesting now, seems fee has something to do with `estimateEncodedSize`, so I don't think we should remove it...

algochoi: got it, sounds reasonable!

michaeldiamant:
For reader awareness:
* `setFee` - Deprecation added in https://github.com/algorand/java-algorand-sdk/pull/85/.
* `setLease` - Deprecated added in https://github.com/algorand/java-algorand-sdk/pull/83.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
   of `TxGroup.assignGroupID(Address address, Transaction ...txns)`.
 * Remove `Account.transactionWithSuggestedFeePerByte` in favor of `Account.setFeeByFeePerByte`.
 * Remove deprecated methods in `Transaction.java`, mark public `Transaction` constructor as hidden in favor of `com.algorand.algosdk.builder.transaction`.
+* Remove deprecated `Transaction.setFee` and `Transaction.setLease` methods
```
michaeldiamant, line 9:
```suggestion
* Remove deprecated `Transaction.setFee` and `Transaction.setLease` methods.
```

michaeldiamant (approved):

PR #443, opened 2022-12-12
Enhancement: Remove v1 algod endpoint

ahangsu:
Implements https://github.com/algorand/algorand-sdk-testing/issues/255

PR #442, opened 2022-12-12
deprecate langspec and logic sig templates

ahangsu:
Remove langspec and dependencies over langspec, remove all logicsig templates:
- https://github.com/algorand/algorand-sdk-testing/issues/254
- https://github.com/algorand/algorand-sdk-testing/issues/223

PR #453, opened 2022-12-15
Changelog entries for breaking changes in V1 to V2

ahangsu:
See #389 and #447, some more entries in `CHANGELOG` need to be added:
- #443 
  - remove v1 endpoint
  - remove `MethodCallParams` public constructor
- #445 
- #442

This should be the unblocker for #447 to be merged into `develop`.


michaeldiamant (approved):
@ahangsu Looks complete - thanks for circling back here!

PR #511, opened 2023-03-01
Fix: improve error message for mismatched args

barnjamin:
Changing the error message to be more accurate

### DIFF
--- a/src/main/java/com/algorand/algosdk/transaction/MethodCallParams.java
+++ b/src/main/java/com/algorand/algosdk/transaction/MethodCallParams.java
@@ -170,7 +170,7 @@ public class MethodCallParams {
                 methodABIts.add(argT.parsedType);
             } else
                 throw new IllegalArgumentException(
-                        "error: the type of method argument is a transaction-type, but no transaction-with-signer provided"
+                        "error: the type of method argument "+i+" is a "+argT.type+", but "+argT.type+" not provided"
                 );
         }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/algorand/algosdk/transaction/MethodCallParams.java
```diff
@@ -170,7 +170,7 @@ public List<TransactionWithSigner> createTransactions() {
                 methodABIts.add(argT.parsedType);
             } else
                 throw new IllegalArgumentException(
-                        "error: the type of method argument is a transaction-type, but no transaction-with-signer provided"
+                        "error: the type of method argument "+i+" is a "+argT.type+", but "+argT.type+" not provided"
```
michaeltchuang, line 173: I think
```suggestion
                        "error: the type of method (" + this.method.name +") argument " + (i+1) + " is a " + this.methodArgs.get(i).getClass().getName() + ", but " + argT.type + " not provided"
```
`argT.type` on both side might be a typo? (or not as helpful)

I tested this and here's my suggestion's output

Screenshot 2023-03-01 at 9 58 24 PM

and here is String as input

Screenshot 2023-03-01 at 9 56 05 PM

barnjamin: nice one, thanks!

winder (approved):
LGTM

PR #555, opened 2023-04-26
api: Regenerate Client Interfaces and implement cucumber tests.

winder:
See commits for where the auto-pr ends and my changes begin.

This uses and updated generator.
generator PR: https://github.com/algorand/generator/pull/61
codegen diff: https://github.com/algorand/java-algorand-sdk/commit/4453df9d9459139166e5438b66ddd74a2be6a063


ahangsu (approved):
looks good on `Simulate*` side

algochoi:
I think we need to regenerate simulate endpoints based on https://github.com/algorand/go-algorand/pull/5335

algochoi:
I think we need to regenerate the models based on https://github.com/algorand/go-algorand/pull/5335

winder:
> I think we need to regenerate the models based on [algorand/go-algorand#5335](https://github.com/algorand/go-algorand/pull/5335)

Thanks, my local algod repo was stale. Regenerated with the current master and it picked up several changes.

algochoi (approved):
Looks good - noting that the simulation endpoints are only generated and not tested.

PR #593, opened 2023-06-13
CICD: Add create-release-pr workflow

algobarb:
## Summary
Add create-release-pr workflow to create release PRs using workflow dispatch.

## Testing
In separate repo:
- Made sure updated Changelog.md looks right
- Made sure README gets updated
- tags are added with a hyphen


scripts/bump_version.py
```diff
@@ -0,0 +1,38 @@
+# This script bumps up the version in setup.py for new releases.
```
excalq, line 1:
```suggestion
# This script bumps up the version in `pom.xml` for new releases.
```

scripts/bump_version.py
```diff
@@ -0,0 +1,38 @@
...
+# Usage: python bump_version.py {new_version} (--read_me <path-to-README.md> --pom_xml <path-to-pom-xml>)
+
+import argparse
+import re
+
+def bump_version(new_version, file_path):
+    with open(file_path, "r") as file:
+        read_me = file.read()
```
excalq, line 9: nit: It's only 'README' for one invocation. Generalize variable name.
```suggestion
        content = file.read()
```

excalq (approved):
Only the comment update and var-name nit, both non-blocking.

scripts/bump_version.py
algobarb: Yes thank you

algobarb: Updated.

Issue #651, opened 2023-10-18
Use of HashSet to filter out duplicates makes the order of foreign arrays unreliable


CiottiGiorgio:
https://github.com/algorand/java-algorand-sdk/blame/f3e6dcdfe24fefb455de52f1d07e1f83b25f6580/src/main/java/com/algorand/algosdk/builder/transaction/MethodCallTransactionBuilder.java#L108

As title. I recently helped a project debug an issue that was appearing at random times. We were able to track it down to the order of the accounts in the foreign account array. I believe this is not the only place that `HashSet` is being used as a filter but it's the only one that we were able to confirm.

CiottiGiorgio:
Thanks!

PR #652, opened 2023-10-24
Patch: Remove de-duping via HashSet from accounts,foriegn apps, and foreign …

gmalouf:
…assets setters in MethodCallTransactionBuilder. Now set solely via lists to ensure stable ordering maintained.

Addresses #651 

Existing unit and integration tests were run / are passing.


winder (approved):
LGTM

Issue #523, opened 2023-03-16
Allow TransactionBuilder to accept `null` for `fee` 


barnjamin:
## Problem

I'd like to rely on suggested params to populate most of the fields but when overriding the `fee` on the transaction with a `flatFee` the builder errors since both `fee` and `flatFee` are set.

If I try to pass `null` to the `fee` method, it only works in the case of `BigInteger` which may be changed at some point.

## Solution

Allow `fee` method in builder to accept `null` for Integer or Long so we can properly unset the fee parameter while still allowing `suggestedParameters` to be passed instead of individual arguments.

subhakundu:
Is someone working on this issue? I would like to work on it if no one is taking a look. This is my first time working on the code base. Please let me know the things I should be keeping in mind.

jannotti:
No one is working on it, but it seems like a nice convenience. Go for it!

subhakundu:
I have a few questions to ask before I raise the PR
1. When we are passing null to BigInteger one ([code](http://tinyurl.com/4rz9vnpu)), we are setting fee as null. I am assuming behaviour will be the same for Long and Integer one as well. In that case, we just need to introduce a null check and set null when null is passed to Long or Integer method. Something like this

```
    public T fee(Integer fee) {
        if (fee != null) {
            if (fee < 0) throw new IllegalArgumentException("fee cannot be a negative value");
            this.fee = BigInteger.valueOf(fee);
        } else {
            this.fee = null;
        }
        return (T) this;
    }
```
2. Also, I am planning to write tests in TestTransaction.java. Because of my lack of experience in code, it would be great if someone could tell me what type of tests I need to perform. I don't want to miss any cases. If there is any other package which is dependent on this codebase, please let me know.

If above approach in (1) looks fine to you, I will raise PR soon. Also, please let me know if you have any feedbacks.

jannotti:
This looks like the right approach to me.  I don't know much about the Java sdk tests, but I'll try to look later. For me, as long as you have a test that sets the fee to null with an Integer and a Long argument, I think you're fine.

Separately, I wonder if another thing we should do is null out `flatFee` whenever `fee` is set and vice versa. @jasonpaulos do you see any reason not to do that?

subhakundu:
Thanks @jannotti for the pointers. Please let me know if I should make flatfee related change in this PR. Or I can create a separate PR for that.

subhakundu:
[...]

PR #660, opened 2023-12-15
DX: Allowing null for fee and flatfee in Long and Integer builder methods.

subhakundu:
This PR contains two changes related to this [issue](https://github.com/algorand/java-algorand-sdk/issues/523).
1. Allowing null for fee and flatfee in Long and Integer builder methods.
2. Adding unit test for testing behaviour of transaction builder if fee and flatfee is null. (#523)

Continuation of discussion [here](https://github.com/algorand/java-algorand-sdk/issues/523#issuecomment-1857197644), I think we do not need to add check to null out flatfee if fee is set and vice versa. The builder logic goes like this ([code](https://github.com/subhakundu/java-algorand-sdk/blob/develop/src/main/java/com/algorand/algosdk/builder/transaction/TransactionBuilder.java#L55))

1. If fee and flatfee are set, we throw an exception. Two values cannot be present in same transaction.
2. If fee and flatfee are null, we set fee to minimum value i.e., 1000 Algos ([code](https://github.com/subhakundu/java-algorand-sdk/blob/develop/src/main/java/com/algorand/algosdk/builder/transaction/TransactionBuilder.java#L59)). I hope I am using right terminologies. Please pardon my mistakes.
3. If fee is not null, we ignore flatfee altogether. If fee is set as null after setting passed fee value, we set fee in the transaction to the minimum value (1000 Algos).
4. If flatfee is null, fee is set to flatfee in the transaction. This is the case where flatfee is used if fee is null.

I have added a unit test to check various scenarios. If not needed, I can remove it. But I believe it will help document different scenarios in better manner. Please let me know your suggestions and concerns.

I have tested the code by running `mvn package` and `mvn clean test` locally.

### DIFF
--- a/src/main/java/com/algorand/algosdk/builder/transaction/TransactionParametersBuilder.java
+++ b/src/main/java/com/algorand/algosdk/builder/transaction/TransactionParametersBuilder.java
@@ -103,8 +103,12 @@ public abstract class TransactionParametersBuilder<T extends TransactionParamete
      * @return This builder.
      */
     public T fee(Integer fee) {
-        if (fee < 0) throw new IllegalArgumentException("fee cannot be a negative value");
-        this.fee = BigInteger.valueOf(fee);
+        if (fee != null) {
+            if (fee < 0) throw new IllegalArgumentException("fee cannot be a negative value");
+            this.fee = BigInteger.valueOf(fee);
+        } else {
+            this.fee = null;
+        }
         return (T) this;
     }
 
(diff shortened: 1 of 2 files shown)


jannotti:
Thank you! I'm willing to take the PR as is (once we figure out why the integration tests fail) because it's better than what we had.

But I did not follow your reasoning about why you don't think we should set `fee` to `null` when `flatFee()` is called with a non-null argument and vice versa.  The original request for this PR was because someone was calling
```
b.suggestedParmams(sp)
// set a flat fee instead of taking the recommended fee per byte
b.flatFee(xxx)
```
and getting a failure because `fee` was still set to a non-null value. `fee` was set by `suggestedParams`, and `flatFee` was set in an attempt to override the suggestion.

So that person had to use
```
b.suggestedParmams(sp)
// set a flat fee instead of taking the recommended fee per byte
b.flatFee(xxx)
b.fee((BigInteger)null) // b.fee((Integer)null) doesn't work, causes NPE
```
They found that untidy, and worried it would change some day.  So now, with your change, they can use `b.fee((Integer)null)` or `b.fee((Long)null)`.  But that's still not very tidy!

I think they really ought to be able to use the original code, that simply calls `flatFee` to override the suggestion.  For that to work, the `flatFee` method must set the `flatFee` field _and_ null out the `fee` field. Otherwise, we are in the situation you labeled `1` and we get an exception.

It seems proper to me that when two fields cannot be used at the same time, they should null the other field if they are set. (Of course, they should _not_ null out the other field if they are set to `null`.)

subhakundu:
Sure. Let me check once again. I will get back with another revision.

PR #714, opened 2024-05-27
Fixes on Account and Transaction

FabioRagno:
Pull request includes two fixes:
1. min-balance attribute added in order to align sdk Account model to the ReST API response
2. removing Transaction.fee initialization at 1000 microAlgos in case of null variable. Fix is necessary for response deserialization in case of transaction submitted with no fees, in particular in case of Atomic Transaction pooled fees. In case of single transaction, if no fee is set, the transaction will be rejected by the chain.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -42,8 +42,8 @@
         <maven.surefire.version>3.0.0-M5</maven.surefire.version>
 
         <!-- Test sources at 1.8 to allow using JUnit5 -->
-        <java.version>1.7</java.version>
-        <java.release.version>7</java.release.version>
+        <java.version>1.8</java.version>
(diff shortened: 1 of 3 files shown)


gmalouf:
@FabioRagno we are closing this specific PR:
- #715 brings the min balance in (this is autogenerated) software and #717 exercises interacting with the new field
- A minimum JDK version upgrade is a major thing - we _probably_ will go to 8 or 11 in the near term, but will be addressed in its own pull request
- The transaction fee initialization fix needs appropriate test cases, if you would like to create appropriate tests (check out the existing ones in the codebase), we can work to accept a PR for this.

Thank you for your contributions!

PR #728, opened 2024-07-25
feat: add in simulate function for ATC and EmptyTransactionSigner

michaeltchuang:
## Summary
 - Ported simulate function to `AtomicTransactionComposer.java` (mirror Go/JS versions)
 - Also created `EmptyTransactionSigner` class

## Testing Plan
 - Added some integration tests to pass `simulate.feature` cucumber tests


src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -326,6 +376,77 @@ public ExecuteResult execute(AlgodClient client, int waitRounds) throws Exceptio
...
+                if (logs == null || logs.isEmpty()) {
+                    throw new Exception("App call transaction did not log a return value");
+                }
+
+                byte[] lastLog = logs.get(logs.size() - 1);
+                if (lastLog.length < 4 || !hasPrefix(lastLog, ABI_RET_HASH)) {
+                    throw new Exception("App call transaction did not log a return value");
+                }
+
+                returnedResult.setRawReturnValue(Arrays.copyOfRange(lastLog, 4, lastLog.length));
```
jasonpaulos, line 428: minor: I would prefer referencing the return hash length instead of hardcoding 4
```suggestion
                returnedResult.setRawReturnValue(Arrays.copyOfRange(lastLog, ABI_RET_HASH.length, lastLog.length));
```

src/main/java/com/algorand/algosdk/transaction/AtomicTransactionComposer.java
```diff
@@ -326,6 +376,77 @@ public ExecuteResult execute(AlgodClient client, int waitRounds) throws Exceptio
...
+     * @param method
+     * @param methodResult
+     * @param pendingTransactionResponse
+     * @return An ABIMethodResult object
+     */
+    public ABIMethodResult parseMethodResponse(Method method, ABIMethodResult methodResult, PendingTransactionResponse pendingTransactionResponse) {
+        ABIMethodResult returnedResult = methodResult;
+        try {
+            returnedResult.setTransactionInfo(pendingTransactionResponse);
+            if (!method.returns.type.equals("void")) {
```
jasonpaulos, line 417:
```suggestion
            if (!method.returns.type.equals(Method.Returns.VoidRetType)) {
```

jasonpaulos (commented):
Thanks for this contribution. I've left comments about improvements I'd like to see, and guidance for how to add support for our existing simulation integration tests.

examples/src/main/java/com/algorand/examples/Reti.java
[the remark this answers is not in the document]
michaeltchuang: @jasonpaulos - Sorry on taking awhile to update this PR, it took me some time to configure my local environment for cucumber tests and getting the necessary free time to work on this.  I had a question though (after I understood your feedback)...does the `Go Algorand SDK` include all the `@simulate` steps in `simulate.feature`?  I was using it as a reference and noticed some missing

I removed the `reti` example test I wrote from this PR since `simulate.feature` test coverage would probably be sufficient

Please also let me know if I'm not on the right path on with the `simulate.feature` tests (i committed the first batch of integration tests into this PR now for review)

PR #868, opened 2025-09-29
Update commons-codec to 1.19.0, addressing CVE reports for 1.12.

gmalouf:


algorandskiy (approved):
Interesting comment about non upgrading to 1.14 but looks like well outdated?

